PR #152, opened 2026-06-07
fix(icons): jitter on save + smart hydration for instant icon render

bohemian-miser:
## Summary

- **Jitter fix** — `react-flow-diagram.tsx`: `layoutsJustArrived` now only triggers on first load (`prevRef === null`) instead of every state update, eliminating icon glitch when autosave fires
- **Smart hydration** — on recipe creation, server synchronously resolves the top icon match per node before saving to Firestore so the UI renders icons on the very first snapshot; remaining `fast_matches` are stored on the node and hydrated client-side after load, with a full search as fallback if none are present

## Test plan

- [x] 222 unit tests passing
- [ ] Create a recipe and verify icons appear immediately on first render (no blank → pop-in delay)
- [ ] Save a recipe and confirm icons don't glitch/reset position when autosave fires
- [ ] Verify fallback: recipe with no fast_matches still gets icons via client-side search


bohemian-miser:
Superseded by #153 and #154

PR #157, opened 2026-06-07
fix(types): remove duplicate local FastMatch type shadowing the import

bohemian-miser:
BatchSearchFn in icon-pipeline.test.ts was using a stale return shape ({embedding, fast_matches}) that diverged from the DataService interface ({icons, matchScores}); updated both.


bohemian-miser:
Superseded by #159 — this fix (and the rest of this branch) is included in the squashed staging commit. Closing in favor of #159.

PR #163, opened 2026-06-10
fix(ai): fail-safe mock-mode guard so prod never serves placeholder icons

### LINKED ISSUE #19: Prod backend function is using placeholder images
@gemini-cli work out why the backed is generating placeholder icons instead of AI generated images

bohemian-miser:
## Root cause (#19)
`lib/ai-service.ts` selected `MockAIService` (which returns `placehold.co` URLs) purely from env flags — `MOCK_AI`, `FUNCTIONS_EMULATOR`, or `NEXT_PUBLIC_USE_FIREBASE_EMULATOR` — with **no production safety guard**. If any of those leaked into a prod runtime (this has happened before), the backend silently served mock icons.

## The guard
Refactored the module-load selection into a pure, testable `selectAIService(env)`. New fail-safe rule: when a mock flag is set **AND** the prod signal is detected, log a loud config-leak error and force `RealAIService` instead of mocks. All other paths (dev, test, emulators, node_cf, real) are byte-for-byte unchanged.

## Prod signal chosen: `NODE_ENV === 'production'`
This is the signal already used consistently across the codebase for prod detection: `lib/firebase-admin.ts`, `app/api/auth/login/route.ts` (secure cookies), and `scripts/prebuild.js`. Next.js production builds and Firebase App Hosting set `NODE_ENV=production` at runtime, while emulators and tests do not — so the guard fires exactly in prod and never disturbs local/test mock usage.

## Prebuild-guard discrepancy note
Memory suggested the MOCK_AI prebuild guard was *absent*; investigation found it **is present** in `scripts/prebuild.js` (lines 8–39) and fails a prod build when `MOCK_AI=true`. However, that is build-time only, keys on `MOCK_AI` alone (not the emulator flags), and does not protect the runtime/cold-start path — so the runtime guard added here is still the real fix. No change made to prebuild.js.

## Verification
- `npm run typecheck` — pass
- `npm run test:unit:pure` — 256 pass / 0 fail (added 6 selection tests, incl. mock+prod→Real and mock+non-prod→Mock)
- eslint on changed files — clean

Closes #19

### DIFF
--- a/recipe-lanes/lib/ai-service.ts
+++ b/recipe-lanes/lib/ai-service.ts
@@ -308,26 +308,49 @@ export class MockAIService implements AIService {
 
 // Default to Real or Mock based on env.
 // Tests can swap this out using setAIService.
-const isMockMode = 
-  process.env.MOCK_AI === 'true' || 
-  process.env.FUNCTIONS_EMULATOR === 'true' || 
-  process.env.NEXT_PUBLIC_USE_FIREBASE_EMULATOR === 'true';
+//
+// Selection is a pure function of env so it can be unit-tested directly
+// (the live `currentService` is resolved once at module-load time below).
+export function selectAIService(env: Record<string, string | undefined> = process.env): AIService {
+    const mockFlagSet =
+        env.MOCK_AI === 'true' ||
+        env.FUNCTIONS_EMULATOR === 'true' ||
+        env.NEXT_PUBLIC_USE_FIREBASE_EMULATOR === 'true';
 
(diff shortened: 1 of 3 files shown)


bohemian-miser:
Superseding this runtime NODE_ENV guard with a structural fix: the mock AI client will be removed from the prod composition/bundle entirely (build-time swap), rather than gated by a runtime flag. Investigating the current wiring first, then strengthening the existing pattern. Closing in favor of that follow-up.

PR #165, opened 2026-06-11
fix(ai): remove MOCK_AI flag — real client always in prod, mock injected via DI

### LINKED ISSUE #19: Prod backend function is using placeholder images

bohemian-miser:
## What

Removes the `MOCK_AI` environment flag entirely and selects the AI client purely via **dependency injection**. This supersedes the earlier "tree-shake the mock by gating on `NODE_ENV`" framing — the previous version pushed to this PR was stale and never reflected the DI approach.

## How

- **`lib/ai-service.ts`** — app/composition code now *always* constructs a real client (`NodeCFAIService` or `RealAIService`). No env flag selects a mock here, and the mock module is never referenced from this file, so it is structurally absent from the production bundle. `getAIService()` / `setAIService()` remain the injection seam.
- **`instrumentation.ts`** (Next.js server-startup hook) — the dev/e2e harness injects `MockAIService` via `setAIService()`, guarded by `NODE_ENV !== 'production'`. Production (`next start`) keeps the real client.
- **`functions/src/index.ts`** — the cloud-functions emulator runs in its own process, so it injects `MockAIService` guarded by `process.env.FUNCTIONS_EMULATOR`. Firebase sets this **only** under the local emulator and never in deployed functions, so prod never loads the mock. A dynamic `require` keeps the mock module out of any prod code path.
- **`tests/helpers/mock-ai.ts`** — `useMockAI()` DI helper; the only way unit tests get the mock.
- **`scripts/prebuild.js`** — the former `MOCK_AI` production guard is removed; there is no flag left that could leak into a build, so there is nothing to guard.

Both injection signals (`NODE_ENV`, `FUNCTIONS_EMULATOR`) are **intrinsic, non-leakable** runtime signals — neither can be accidentally set in a deploy config the way a custom `MOCK_AI` env var could (cf. the earlier MOCK_AI-to-prod incident).

## Proof

- `grep -rn "process.env.MOCK_AI" recipe-lanes` → only matches are inside `verify-production-logic.test.ts` (which *asserts* zero references elsewhere). No functional reads remain anywhere in app/composition/config code.
- `npm run typecheck` ✅
- `npm run test:unit:pure` → 261 pass / 1 fail; the single failure is the pre-existing `hybrid-plumbing` legacy-network test (untouched by this branch).
- `npm run build` ✅, then `grep -rl "MockAIService" .next/server .next/static` (excl. maps) → **nothing**. The mock is not in the production bundle.
- Functions emulator DI verified at runtime: under `FUNCTIONS_EMULATOR=true`, `getAIService()` returns a `MockAIService` instance.

Closes #19


bohemian-miser:
i don't understand the point of this cl, you don't actually remove process.env.MOCK_AI ??

bohemian-miser:
Heads up: the version previously pushed to this PR was **stale** — it still showed the old "tree-shake the mock via `NODE_ENV` gating" approach. I've rebased onto current `main` and force-pushed the real change (`757bd37`).

**`process.env.MOCK_AI` is now fully removed.** Zero-grep proof:

```
$ grep -rn "process.env.MOCK_AI" recipe-lanes
recipe-lanes/tests/verify-production-logic.test.ts:49:  it('no process.env.MOCK_AI reference exists anywhere in source/config', ...)
recipe-lanes/tests/verify-production-logic.test.ts:58:          '"process.env.MOCK_AI" ' +
recipe-lanes/tests/verify-production-logic.test.ts:67:    assert.strictEqual(out.trim(), '', ...)
```

The only matches are inside `verify-production-logic.test.ts`, which *asserts* there are zero references anywhere else (it excludes itself). No app/composition/config code reads the flag.

**The mock is now injected via pure DI**, never selected by an env flag:
- **dev / e2e** — `instrumentation.ts` (Next server-startup hook) injects `MockAIService` via `setAIService()`, guarded by `NODE_ENV !== 'production'`.
- **unit tests** — `tests/helpers/mock-ai.ts` `useMockAI()` helper calls `setAIService()`.
- **functions emulator** — `functions/src/index.ts` injects the mock guarded by `FUNCTIONS_EMULATOR` (Firebase sets this only under the local emulator, never in deployed functions). Verified at runtime: `getAIService()` returns a `MockAIService` instance when that var is set.

Prod has **zero** mock path: `npm run build` then `grep -rl "MockAIService" .next/server .next/static` (excl. maps) returns nothing. Both injection signals (`NODE_ENV`, `FUNCTIONS_EMULATOR`) are intrinsic and non-leakable, unlike the old custom `MOCK_AI` var.

Issue #172, opened 2026-06-12
hide icon_overview behind auth


bohemian-miser:

bohemian-miser:
still need to hide the icon generation behind auth

PR #176, opened 2026-06-13
fix(icon_overview): public view, login-gated forge, admin-gated controls (#172)

bohemian-miser:
## Bug 172 — rescope auth on `/icon_overview`

**Note:** the bug title says "hide behind auth", but there was no real view-gate — the page already rendered for logged-out users. The actual defect was that the **forge action had no auth check at all**. This PR fixes the real problem and matches the agreed product spec.

### Behavior
- **Anyone can VIEW** `/icon_overview` (including logged-out visitors).
- **Forging is gated** server-side in `addIngredientNodeAction`.
- **Admin management controls** (Clear Queue, Retry) are hidden from non-admins in the UI, matching the existing server-side `isAdmin` checks.

### Tests
- New "Forge Auth Scoping" cases in `tests/admin-security.test.ts`.

First in a stack of 3 (→ abuse controls → alerting).

### DIFF
--- a/recipe-lanes/app/actions.ts
+++ b/recipe-lanes/app/actions.ts
@@ -89,6 +89,12 @@ export async function createDebugRecipeAction() {
 }
 // icon_overview and tests only.
 export async function addIngredientNodeAction(recipeId: string, ingredientName: string) {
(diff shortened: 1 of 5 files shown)


bohemian-miser (commented):
LGTM

PR #177, opened 2026-06-13
feat(icon-queue): runtime config doc + abuse controls

bohemian-miser:
## Runtime queue config + abuse controls

Adds a single Firestore `config/icon_queue` doc read by both the Cloud Function and the admin UI, so abuse controls are adjustable **at runtime with no redeploy**.

### Controls
- **`paused`** (default false) — `processIconTask` checks it first; when paused, re-enqueues with backoff so work auto-resumes on un-pause (no dropped/failed tasks).
- **`allowAnonForge`** (default true) — when false, anonymous forge attempts are rejected with a login-prompt-friendly error.
- **`perUserDailyCap`** (default 100) — per-user daily forge ceiling, counted via `icon_forge_usage/{uid}_{utcDay}` (one read + atomic increment; queue docs are deleted on processing so they can't be counted).
- **Admin-only config panel** on `/icon_overview`, gated server-side and in the UI.

Cloud Tasks concurrency (`maxConcurrentDispatches`/`maxDispatchesPerSecond`) intentionally **stays in GCP**, not the app.

### Tests
- New `tests/icon-queue-config.test.ts` + extended forge auth cases. Verified 14/14 against emulators.

**Stacked on #172** — review/merge that first. Base will retarget to `main` once #172 lands.

### DIFF
--- a/recipe-lanes/app/actions.ts
+++ b/recipe-lanes/app/actions.ts
@@ -59,6 +59,11 @@ export async function forgeIconAction(recipeId: string, ingredientName: string, 
     try {
         const session = await getAuthService().verifyAuth();
         const userId = session?.uid;
+        const forgeBlock = await checkForgeAllowed(userId);
+        if (forgeBlock) return { success: false, error: forgeBlock };
+        if (userId) incrementUserForgeCount(userId).catch(e =>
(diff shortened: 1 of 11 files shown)


bohemian-miser (commented):
LGTM

PR #200, opened 2026-07-04
Default anon-created recipes to public visibility

### LINKED ISSUE #151: Anon default to public (+ vetting)

bohemian-miser:
## What & why

Closes #151. Recipes saved by anonymous (not-signed-in) users should default to `public` visibility instead of `unlisted`, gated by the existing vetting flow (`isVetted`) before appearing in the gallery — that vetting half was already implemented on `main`; this closes the remaining gap.

Changes:
1. `finalizeAndSaveRecipe` in `recipe-lanes/app/actions.ts` no longer hardcodes `visibility = 'unlisted'` for new/forked recipes — it now leaves visibility unset so `saveRecipe` applies an auth-based default (anon → `public`, signed-in → `unlisted`). Updates to an existing recipe you own still preserve the recipe's current stored visibility, unchanged from before.
2. `saveRecipeAction`'s `visibility` parameter is now optional (was defaulted to `'unlisted'`).
3. Both `DataService.saveRecipe` implementations (`FirebaseDataService`, `MemoryDataService`) in `recipe-lanes/lib/data-service.ts` now only write/change `visibility` on an **update** when the caller explicitly passes a value — otherwise the stored value is preserved. On **create**, they default to `public` for anonymous saves and `unlisted` for signed-in saves (unless the caller passes an explicit value).

This also fixes a pre-existing bug found during scoping: because `visibility` defaulted to `'unlisted'` in both the action and the data-service layer, and most autosave call sites in `app/lanes/page.tsx` don't pass `visibility` explicitly, *every* autosave of an existing recipe was silently resetting its visibility back to `unlisted` — including recipes a user had explicitly made `public`. That clobber is fixed as part of this change (updates now only touch visibility when explicitly requested).

No gallery or Firestore-rules changes were needed — `getPublicRecipes`/`searchPublicRecipes` already gate on `visibility == 'public' AND isVetted == true`, and `isVetted` already defaults to `false` on every new recipe doc.

Smallest change that closes the issue — no changes to the god files (`data-service.ts` changes are limited to the `saveRecipe` default-visibility logic; `app/lanes/page.tsx` and `react-flow-diagram.tsx` untouched).

## How verified

- Local: `npm run lint` — 0 errors (pre-existing warnings only, none new).
- Local: `npm run typecheck` (`tsc --noEmit`) — passes cleanly.
- Local: `npm run test:unit:pure` — 307/307 tests pass (run both standalone and again via the pre-commit hook).
- Reviewed all existing test call sites of `saveRecipe`/`saveRecipeAction` (`tests/data.test.ts`, `tests/social-features.test.ts`, `tests/data-helpers-transaction.test.ts`, `tests/impression-rejection.test.ts`) — all pass `visibility` explicitly or construct Firestore docs directly, so none depend on the old implicit-default behavior.
- CI: GitHub Actions run on this PR (link/status to follow once available).

## Risks / unsure about

- **Behavior change for signed-in autosaves too**: previously *every* save (including autosaves of existing docs) reset visibility to `unlisted` regardless of what it was set to. After this change, an existing recipe's visibility is preserved across autosaves unless a caller explicitly changes it. I believe this is strictly a bug fix (matches the scoping analysis in the issue thread), but it's a behavior change beyond the anon-only ask, so flagging for a careful look.
- I did not start the Firebase emulators or run integration/e2e tests locally (per this run's local-check-only constraints) — only lint/typecheck/pure unit tests were run. CI's emulator-backed integration suite is the verification source of truth for the Firestore-backed `saveRecipe` path.
- Did not verify in a running browser/UI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HMuDzj1vttSWBqBhtaCVcJ)_

### DIFF
--- a/recipe-lanes/app/actions.ts
+++ b/recipe-lanes/app/actions.ts
@@ -181,7 +181,7 @@ export async function addIngredientNodeAction(recipeId: string, ingredientName: 
 
 
 // New Action for "Optimistic Return + Background Trigger"
-export async function createVisualRecipeAction(recipeText: string, currentId?: string): Promise<{ id?: string; error?: string }> {
+export async function createVisualRecipeAction(recipeText: string, currentId?: string, claimToken?: string): Promise<{ id?: string; error?: string }> {
     try {
         console.log('[createVisualRecipeAction] 🚀 Starting...');
 
@@ -206,7 +206,7 @@ export async function createVisualRecipeAction(recipeText: string, currentId?: s
             graph.serves = 1;
         }
 
-        return await finalizeAndSaveRecipe(graph, currentId);
+        return await finalizeAndSaveRecipe(graph, currentId, claimToken);
     } catch (e: any) {
         console.error('[createVisualRecipeAction] Failed:', e);
         return { error: e.message || 'Failed to process recipe.' };
@@ -219,7 +219,7 @@ export async function createVisualRecipeAction(recipeText: string, currentId?: s
  * exact same icon pre-population / save / forge pipeline as the text flow. The
  * image is processed in-memory only — never persisted.
  */
-export async function createVisualRecipeFromImageAction(imageDataUrl: string, currentId?: string): Promise<{ id?: string; error?: string }> {
+export async function createVisualRecipeFromImageAction(imageDataUrl: string, currentId?: string, claimToken?: string): Promise<{ id?: string; error?: string }> {
     try {
         console.log('[createVisualRecipeFromImageAction] 📷 Starting...');
 
@@ -242,7 +242,7 @@ export async function createVisualRecipeFromImageAction(imageDataUrl: string, cu
         graph.baseServes = parsedServes;
         graph.serves = parsedServes;
 
-        return await finalizeAndSaveRecipe(graph, currentId);
+        return await finalizeAndSaveRecipe(graph, currentId, claimToken);
     } catch (e: any) {
         console.error('[createVisualRecipeFromImageAction] Failed:', e);
         return { error: e.message || 'Failed to process recipe photo.' };
@@ -256,7 +256,7 @@ export async function createVisualRecipeFromImageAction(imageDataUrl: string, cu
  * gated background icon forging. Expects `graph` to already have title /
  * serves / originalText populated by the caller.
  */
-async function finalizeAndSaveRecipe(graph: RecipeGraph, currentId?: string): Promise<{ id?: string; error?: string }> {
+async function finalizeAndSaveRecipe(graph: RecipeGraph, currentId?: string, claimToken?: string): Promise<{ id?: string; error?: string }> {
     try {
         // 2. Synchronous icon pre-population — hydrate the top match per node before saving
         // so the UI renders with icons on first snapshot, then the client hydrates the rest.
@@ -342,7 +342,7 @@ async function finalizeAndSaveRecipe(graph: RecipeGraph, currentId?: string): Pr
         }
         
         console.log('[createVisualRecipeAction] 💾 Saving initial recipe...');
-        const id = await getDataService().saveRecipe(graph, targetId, userId, visibility);
+        const id = await getDataService().saveRecipe(graph, targetId, userId, visibility, undefined, claimToken);
 
         // Return the ID immediately — the client's snapshot listener will pick up
         // icon shortlists as resolveRecipeIcons writes them to Firestore.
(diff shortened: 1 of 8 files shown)


bohemian-miser:
you didn't add any tests! make some comprehensive tests that cover all the cases like copies and copies of copies etc and if an anon recipe is made public set the owner as Anon and make it so that authenticated people cant take ownership of it and add tests for this too

bohemian-miser:
Addressed in the latest commit:

- **Ownership takeover fixed**: `saveRecipe` (both `FirebaseDataService` and `MemoryDataService`) now strips `ownerId`/`ownerName` from any write to a doc that currently has no owner, so an anon-created recipe can never be silently claimed by a signed-in user's save. Ownership stays anon permanently once created that way. The sanctioned way to take ownership of an anon recipe's content is an explicit copy (`copyRecipe`), which is unaffected and still gives the copier ownership of their new copy.
- **Anon owner label**: anon-created public recipes now get `ownerName = 'Anon'` at creation time (also fixed a `MemoryDataService.getRecipe` bug that was fabricating `"User <id>"` instead of returning the real stored `ownerName`, which was hiding this from tests).
- **Tests added** (12 new, all passing locally — 319/319 total on `test:unit:pure`):
  - Copy of a copy: each copy independently owned by its own copier, chain doesn't mutate ancestors.
  - Copying an anon-owned public recipe: the copier legitimately becomes owner of their copy; the anon original stays unowned.
  - Anon create → `public` + `ownerName: 'Anon'`; signed-in create → `unlisted`.
  - Autosave omitting `visibility` preserves the stored value (the original clobber fix).
  - A signed-in save on an anon-owned recipe does **not** grant ownership (title edit still saves, ownership doesn't transfer).
  - An anon save attempt on someone else's recipe is still rejected.
  - Forking (`createVisualRecipeAction` with `currentId`) someone else's recipe, including an anon-owned one, creates a new doc without touching the original's ownership; forking a fork chains `sourceId` correctly (copy-of-a-copy at the action layer).

Pushed to the same branch (`fix/issue-151-anon-public-vetting`) as commit `a9a5109` — note the branch was rebased onto latest `main` in between (GitHub auto-updated it), so the PR now shows 2 commits instead of an amended 1.

---

bohemian-miser:
lets make it that when an anon recipe is made it puts a token in the browser in local storage and we hash that and put that in the recipe and a logged in account can only claim it if they provide the token that hashes to the value stored in the anon recipe. Make sure this only works for recipes that already have no owner so a recipe takeover is not possible

bohemian-miser:
Implemented the claim-token idea, folded into the existing save flow (no separate button/action):

- At anon creation (`createVisualRecipeAction`/`createVisualRecipeFromImageAction`), the browser mints a random token via `crypto.randomUUID()`, sends it once so `saveRecipe` stores only its SHA-256 hash (`claimTokenHash`) on the doc, and keeps the plaintext in `localStorage` under `claim_token_<recipeId>`.
- The existing save path (`useSaveAndFork.performSave`, used by the normal Save/Share actions) automatically attaches that token for the current recipe id if one exists — no new UI, no separate claim step. `saveRecipeAction` threads it to `dataService.saveRecipe`.
- `saveRecipe` only transfers ownership when: the recipe currently has **no owner**, a token was presented, and it hashes to the stored value. Any other case (no token, wrong token, already claimed, already owned by someone else) falls through to the existing anti-takeover behavior from the last round — the save still goes through, ownership just doesn't move.
- Token is cleared from `localStorage` after one attempt (success or not), and cleared server-side (`claimTokenHash` deleted) once actually claimed, so it can't be reused or handed to someone else.

Tests added in `tests/social-features.test.ts` (5 new, 324/324 passing locally on `test:unit:pure`): successful claim with the correct token, wrong token leaves it unowned, a recipe created without a token is unclaimable, a spent token can't be reused by a second claimant, and presenting a token against an already-owned (non-anon) recipe is a safe no-op.

Pushed as commit `6b2d338` on the same branch/PR.

---

bohemian-miser:
could you have consolidated the claim code to save or finalizeAndSave ? seems like we don't need it in two places. I'd really like to keep this simple. maybe it's worth putting in it's own lil lib in case we decide to change how this works i wouldn't want to have to change multiple locations, but that might be overkill

bohemian-miser:
Done in `e5ff057` — went with the "lil lib" option, since the real duplication wasn't in the actions (those just thread a parameter through) but in the two `DataService.saveRecipe` implementations, which each spelled out the hash/verify rules, plus the `claim_token_<id>` localStorage key being written in `page.tsx` and read in `useSaveAndFork`.

Now:
- `lib/recipe-lanes/claim-token.ts` — the whole server-side policy in one place: `hashClaimToken`, `claimHashForCreate` (anon creations stamp a hash), `isValidClaim` (signed-in + exact token + stored hash). Both `saveRecipe` implementations are one-line calls into it.
- `lib/recipe-lanes/claim-token-client.ts` — the browser half: `mintClaimToken` and the storage key (mint/store/get/clear), mirroring the existing `draft-persistence.ts` pattern. `page.tsx` and `useSaveAndFork` call these instead of touching `localStorage` keys directly.

So if you change how claiming works later, it's one file for the server rules and one for the client storage. No behavior change — the existing end-to-end claim tests pass unchanged, and I added direct unit coverage for both lib halves (22/22 in `social-features.test.ts`, 321/321 pure suite locally via the pre-commit hook).

Also FYI: the earlier e2e failure on this PR ("Persistence Regressions: Issue 74 & 61") was an unrelated onSnapshot-timing flake — the test's background write goes through the admin SDK, not `saveRecipe`, and the e2e user is signed in so no claim token is ever minted in those flows. It passed on re-run with no code change; all checks were green on `6b2d338` before this refactor kicked off a fresh CI round.

---

PR #223, opened 2026-07-05
ci: add Claude @claude-mention comment responder

bohemian-miser:
## Summary
- Adds a repo-wide GitHub Actions workflow (`.github/workflows/claude.yml`) using `anthropics/claude-code-action@v1`, triggered by `issue_comment` and `pull_request_review_comment` events.
- Commenting `@claude <instruction>` on any PR or issue will trigger Claude to respond/act.
- Guarded so it only fires on `@claude` mentions from non-bot commenters (`github.event.comment.user.type != 'Bot' && contains(github.event.comment.body, '@claude')`).

## Setup required before this works
This workflow is **inert** until the owner completes two manual steps:
1. Install the Claude GitHub App: https://github.com/apps/claude
2. Add an `ANTHROPIC_API_KEY` repo secret (or swap the workflow to use a `claude_code_oauth_token` secret instead, to bill against the Claude subscription rather than API credits).

## Test plan
- [x] Validated YAML parses via `python3 -c "import yaml; yaml.safe_load(...)"`
- [ ] After merge + owner setup, comment `@claude` on a test PR/issue and confirm the workflow triggers


bohemian-miser:
don't want to use api

PR #225, opened 2026-07-05
fix(lanes): keep carrot/pan legend visible while any icon is still loading (#60)

### LINKED ISSUE #60: Show carrot and pan legend while there are still carrots and pans
@gemini-cli /invoke
So long as there's an icon that hasn't loaded, the legend should show Show carrot and pan.

bohemian-miser:
Closes #60.

## What & why

The lanes legend's "🥕 Ingredients / 🍳 Actions" explainer tells the user what the placeholder carrot/pan icons mean while real icons are being generated. It was gated on:

```ts
const hasIcons = graph?.nodes.some(n => hasNodeIcon(n));   // ...{!hasIcons && <legend/>}
```

`hasIcons` is true as soon as **any one** node resolves an icon, so the explainer vanished the moment the *first* icon loaded — even though other nodes were still showing placeholder carrots/pans. Issue #60: *"So long as there's an icon that hasn't loaded, the legend should show carrot and pan."*

## The change (smallest that closes it)

- **`lib/recipe-lanes/model-utils.ts`** — new pure helper `hasPendingIcons(graph)`: returns `true` while **any** node still lacks a resolved icon (`!hasNodeIcon(n)`), and `false` for a null/empty graph (no placeholders to explain). This is the exact inverse condition the legend needs.
- **`app/lanes/page.tsx`** — replace `hasIcons`/`!hasIcons` with `const iconsPending = hasPendingIcons(graph ?? null)` and render both the desktop and mobile legends when `iconsPending`. The explainer now stays up until **every** node has an icon, then hides.

The god-file (`app/lanes/page.tsx`) edit is confined to the one derived flag and the two `{iconsPending && …}` conditions — no logic beyond swapping the predicate. All decision logic lives in the pure, unit-tested helper.

## How verified

- **New pure unit tests** — `tests/model-utils.test.ts`, suite `hasPendingIcons (issue #60 …)` (tier: `test:unit:pure`, runs in the CI **fast-checks** job). Cases:
  - null / empty graph → `false`
  - no node has an icon yet (all placeholders) → `true`
  - **some** icons still unloaded (the regression: one resolved + one placeholder) → `true` — this is the exact scenario the old `hasIcons` got wrong
  - every node has a resolved icon → `false`
- **Local gate (pre-commit hook), all green:** `npm run lint` (0 errors, warnings are pre-existing repo-wide), `npm run typecheck` (clean), and the full `npm run test:unit:pure` — **330/330 pass**, including the 4 new cases.
- The `integration` and `e2e` suites run unchanged as regression guards (the change touches `app/`, `lib/`, `tests/`, so CI is not docs-only-skipped).

## Risks / unsure about

- **"Loaded" = model-level, not pixel-level.** `hasNodeIcon` reflects whether a node has a resolved icon in its shortlist, not whether the `` has finished painting in the browser. That matches the placeholder-render condition (the node shows the carrot/pan emoji exactly when it has no resolved icon URL), so it's the right signal for this legend.
- **No component/e2e test for the rendered legend visibility** — the conditional lives in the `app/lanes/page.tsx` client god-file; the branch it now keys on (`hasPendingIcons`) is covered at the pure-unit tier. An e2e assertion on the legend node could be added if desired.

---
_Generated by [Claude Code](https://claude.ai/code/session_01M6NpXn7YLKki44ikrwdpte)_

### DIFF
--- a/recipe-lanes/app/lanes/page.tsx
+++ b/recipe-lanes/app/lanes/page.tsx
@@ -32,7 +32,7 @@ import { standardizeIngredientName, formatDisplayName } from '@/lib/utils';
 import { IngredientsSidebar } from '@/components/recipe-lanes/ui/ingredients-sidebar';
 import { TimelineView } from '@/components/recipe-lanes/timeline-view';
 import type { RecipeGraph, LayoutModeId } from '@/lib/recipe-lanes/types';
-import { hasNodeIcon, preserveNodeShortlist, getNodeShortlistLength, getNodeIngredientName, getNodeHydeQueries, extractBatchIngredients } from '@/lib/recipe-lanes/model-utils';
+import { hasPendingIcons, preserveNodeShortlist, getNodeShortlistLength, getNodeIngredientName, getNodeHydeQueries, extractBatchIngredients } from '@/lib/recipe-lanes/model-utils';
 import { useRecipeStore } from '@/lib/stores/recipe-store';
 import { LayoutMode } from '@/lib/recipe-lanes/layout';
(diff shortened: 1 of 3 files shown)


bohemian-miser:
this is not needed the icon thing has been revamped since this bug was made

Issue #155, opened 2026-06-07
smaller leaf nodes as a global setting


bohemian-miser:

bohemian-miser:
🤖 Triage analysis (automated):

**Restated intent:** Add a global setting to render leaf nodes smaller.

**Likely area:** Node sizing in `recipe-lanes/lib/recipe-lanes/layout.ts` and node components under `recipe-lanes/components/recipe-lanes/nodes/` (minimal-node*.tsx, timeline-node.tsx); global settings likely belong in `recipe-lanes/lib/stores/recipe-store.ts` or `lib/config.ts`.

**Suspected cause / approach:** No leaf-specific sizing exists; would need a leaf detection (out-degree 0) plus a persisted global toggle feeding node dimensions into layout + node render.

**Effort:** M **Value:** L

bohemian-miser:
Opened #236 for this — a global **"Small leaves"** toggle in the diagram toolbar that renders leaf nodes (out-degree 0: final dishes / unused ingredients) at 70% scale.

Core leaf-detection logic is a new pure helper `getLeafNodeIds` (covered by `tests/leaf-nodes.test.ts`); the setting lives in the store as an in-memory UI pref (like `iconStyle`), kept out of `mergeSnapshot`. Watching CI on #236.

---
_Generated by [Claude Code](https://claude.ai/code)_

bohemian-miser:
Opened PR #237 for this: a global **Leaves: Normal / Small** toolbar toggle that renders leaf nodes (out-degree 0 — terminal steps nothing else consumes) smaller. Includes pure unit tests for the leaf detection and the store toggle.

One note for review: I interpreted "leaf" as **out-degree 0** (terminal outputs, per the triage on this issue). If you actually meant the raw source ingredients (in-degree 0), it's a one-line flip — let me know.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #226, opened 2026-07-05
fix: guard against non-finite React Flow handle coords in edge geometry (#30)

### LINKED ISSUE #30: Graph arrows point to incorrect coordinates after save/reload
Arrows sometimes detach or point to incorrect spots on the nodes. This typically happens after saving and reloading the graph, though not consistently. Needs investigation into handle positioning and edge coordinate persistence.

bohemian-miser:
Closes #30.

## What & why

Issue #30: graph arrows detach or point to incorrect coordinates after save/reload — intermittently. The owner's triage narrowed it to the edge-geometry helpers in `lib/recipe-lanes/graph-utils.ts` misbehaving during hydration, before React Flow has measured node/handle bounds.

Root cause (traced): the floating edge (`components/recipe-lanes/edges/floating-edge.tsx`) passes React Flow's `sourceX/sourceY/targetX/targetY` into `getEdgeParams` as handle positions. Before RF measures a node's handle bounds after a reload, those coordinates can be **`NaN`** (undefined-bounds arithmetic). The guards were `typeof sourceX === 'number'` — but `typeof NaN === 'number'` is `true`, so `NaN` slipped through. It then propagated through `getCenter` → `getBBox`/`getNodeIntersection` into **`NaN` edge endpoints**, which render as arrows detached from their nodes or pointing at (0,0)-ish garbage. Once RF re-measures and re-renders, coordinates self-correct — hence the "intermittent, after reload" symptom.

## The fix (smallest change)

- **`lib/recipe-lanes/graph-utils.ts`** — new pure exported guard `isFiniteHandlePos(pos)` (`!!pos && Number.isFinite(x) && Number.isFinite(y)`). `getEdgeParams` (the sole public entry) now sanitizes both handle positions up front, so a non-finite handle is treated as "no handle position" and falls back to the existing node-geometry-derived centers instead of poisoning `getCenter`/`getBBox`/`getRadius`. Sanitizing at this one point covers all three.
- **`components/recipe-lanes/edges/floating-edge.tsx`** — replaces the weak `typeof` checks with `isFiniteHandlePos`, rejecting `NaN`/`Infinity` at the source (defense in depth).

No node-geometry constants or the reference-preserving store logic were touched; the valid-handle path is unchanged, so correctly-measured edges render exactly as before. `0` is intentionally still accepted (a node/handle legitimately at the origin is finite).

## How verified

Added to `tests/graph.test.ts` (pure tier — already registered in the `test:unit:pure` list, runs in the CI **fast-checks** job):
- **`returns finite endpoints when handle positions are NaN (hydration)`** — the direct regression: `getEdgeParams(n1, n2, {x:NaN,y:NaN}, {x:NaN,y:NaN})` must yield finite `sx/sy/tx/ty`. This test **fails on the pre-fix math** (I confirmed it produces `NaN` before the sanitization) and passes after.
- **`treats non-finite handle positions the same as no handle position`** — `NaN`/`Infinity` handles produce exactly the no-handle fallback coordinates.
- **`still honors a valid (finite) handle position`** — finite handles are not discarded.
- **`isFiniteHandlePos`** — accepts finite pairs (incl. `0` and negatives); rejects `undefined`/`null`/`NaN`/`Infinity`.

Local gate: `npm run lint` (0 errors), `npm run typecheck` (clean), scoped `tests/graph.test.ts` → 9/9 pass. The pre-commit hook (lint + typecheck + full `test:unit:pure`) ran green on commit. CI's `fast-checks`, `integration`, and `e2e` jobs execute for this change (it touches `lib/`, `components/`, `tests/` — not docs-only).

## Risks / unsure about

- This eliminates the `NaN`-endpoint ("detached arrow") class deterministically. If any residual after-reload misalignment is instead caused by RF supplying a *finite-but-stale* `{0,0}` before measurement, that can't be distinguished from a legitimate origin handle by value alone and is out of scope here — but the reachable, demonstrable defect (NaN through a `typeof` guard) is fixed.
- The non-minimal `node.width ?? 100` fallback in `getCenter`/`getRadius` is effectively dead code (all real recipe nodes are `type: 'minimal'`; lanes carry no edges; timeline nodes use their own edge geometry), so it was left untouched to keep the change minimal.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gq7p2M7bCcbFrpV3N4J8ci)_

### DIFF
--- a/recipe-lanes/components/recipe-lanes/edges/floating-edge.tsx
+++ b/recipe-lanes/components/recipe-lanes/edges/floating-edge.tsx
@@ -31,7 +31,7 @@ interface EdgeProps {
   style?: React.CSSProperties;
   data?: any;
 }
-import { getEdgeParams } from '../../../lib/recipe-lanes/graph-utils';
+import { getEdgeParams, isFiniteHandlePos } from '../../../lib/recipe-lanes/graph-utils';
 import { useRecipeStore } from '../../../lib/stores/recipe-store';
 
 function FloatingEdge({ id, source, target, markerEnd, style, data, sourceX, sourceY, targetX, targetY }: EdgeProps) {
@@ -45,11 +45,16 @@ function FloatingEdge({ id, source, target, markerEnd, style, data, sourceX, sou
     return null;
   }
 
+  // Only forward handle positions once RF has actually measured them. Before
+  // then RF can supply NaN coords, and typeof NaN === 'number' would sneak them
+  // past a plain typeof check into the edge geometry (issue #30).
+  const sourceHandle = { x: sourceX, y: sourceY };
+  const targetHandle = { x: targetX, y: targetY };
   const { sx, sy, tx, ty, sourcePos, targetPos } = getEdgeParams(
       sourceNode,
       targetNode,
-      (typeof sourceX === 'number' && typeof sourceY === 'number') ? { x: sourceX, y: sourceY } : undefined,
-      (typeof targetX === 'number' && typeof targetY === 'number') ? { x: targetX, y: targetY } : undefined,
+      isFiniteHandlePos(sourceHandle) ? sourceHandle : undefined,
+      isFiniteHandlePos(targetHandle) ? targetHandle : undefined,
       {
         source: sourceNode.data?.isLeaf ? leafNodeScale : 1,
         target: targetNode.data?.isLeaf ? leafNodeScale : 1,
--- a/recipe-lanes/lib/recipe-lanes/graph-utils.ts
+++ b/recipe-lanes/lib/recipe-lanes/graph-utils.ts
@@ -51,6 +51,21 @@ function getFrame(node: Node, handlePos?: { x: number, y: number }, scale = 1): 
     return getClassicFrame(handlePos, isIngredient, scale);
 }
 
+/**
+ * A handle position handed to us by React Flow is only usable when both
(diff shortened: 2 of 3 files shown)


bohemian-miser:
resolve conflicts pls

bohemian-miser:
Done — rebased onto latest `main` and resolved the conflicts (`4229f4e`, force-pushed as `824f55d`).

Both conflicts were in files main refactored in parallel (the `edge-anchors.ts` extraction + leaf-node `scale`/`scales` params from #155/#156). I combined both changes rather than dropping either:
- **`graph-utils.ts` `getEdgeParams`** — the non-finite handle sanitization now feeds the sanitized handles into main's scale-aware `getCenter`/`getBBox`/`getRadius` (and I dropped my obsolete `!!handle` arg in favor of main's numeric `sScale`/`tScale`).
- **`floating-edge.tsx`** — kept main's `useRecipeStore` leaf-scale import and the `scales` third argument, with the handle args now guarded by `isFiniteHandlePos`.

Local re-check on the merged tree: typecheck clean, lint 0 errors, `tests/graph.test.ts` 9/9 pass. Watching CI on the new commit.

---

bohemian-miser:
Closing this as I can no longer reproduce the bug.

PR #235, opened 2026-07-05
fix(lanes): keep recipe title visible in mobile header

### LINKED ISSUE #139: can't see title on mobile

bohemian-miser:
Closes #139

## What & why

On a phone the recipe title in the `/lanes` header was invisible. The header is a single 56px row using `justify-between`: logo + editable title on the left, a nav/action cluster on the right. Both sides default to `flex: 0 1 auto`, and the right cluster is **wider than a phone viewport** (measured ~387px on iPhone 12's 390px width). Because a longer title gives the left side a larger flex-basis, the left side absorbed most of the negative free space and the title container collapsed to **0px** — the title vanished. A previous `truncate`-only tweak couldn't help because the container itself was being squeezed to zero.

The owner's triage suspected the missing `shrink-0` on the nav cluster, but I verified (headless-Chromium layout measurement) that `shrink-0` alone is insufficient: the nav simply doesn't fit alongside a title on a 390px screen, so it would overflow and still leave the title at 0px. The nav footprint has to shrink on mobile.

## What changed (all in `app/lanes/page.tsx`, all `md:`-gated so desktop is untouched)

- Title container gets flex priority: `flex-1 min-w-0` (claims remaining space, truncates instead of collapsing).
- Nav cluster is `shrink-0` so it never steals the title's space, with tighter mobile spacing (`gap-1 md:gap-2`, `px-2 md:px-3`).
- Hide the always-on "Gallery" text label on mobile (`hidden md:inline`) — matches the existing Mine/Starred/New labels.
- Hide the two secondary "Feedback" and "GitHub" links on mobile (`hidden md:flex`).

Net effect measured on a 390px viewport: the title container goes from **0px → ~97px** of visible (truncating) width. The header height stays 56px, so the `top-14`-anchored ingredients sidebar and other overlays are unaffected. All four primary nav items (Gallery, Mine, Starred, New) remain reachable on mobile.

## How verified

- **New e2e regression test** `e2e/mobile-header.spec.ts` → *"long recipe title stays visible on a phone viewport"*: sets the iPhone-12 viewport, creates a recipe, seeds a deliberately long title via an admin write (reaches the editor through onSnapshot, same mechanism `graph.spec` uses), then asserts the rendered `<h1>` is visible and its bounding-box width is > 40px. Before the fix this width was ~0 (test fails); after, it truncates but stays readable (test passes).
- **Local pre-checks**: `npm run lint` (0 errors), `npm run typecheck` (clean), `npm run test:unit:pure` (**340/340 pass**) — all green via the pre-commit hook.
- CI (`fast-checks`, `integration`, `e2e`) still to confirm on this PR.

## Risks / unsure about

- Product judgement: Feedback + GitHub links are hidden on phones (they remain on ≥`md`). This is the smallest reduction that guarantees real title width without changing header height; if you'd rather keep them, an overflow/hamburger menu would be the alternative but is a larger change.
- The e2e width threshold (40px) is deliberately conservative (well below the measured ~97px) to avoid flakiness across font metrics while still failing hard on the 0px regression.

---
_Generated by [Claude Code](https://claude.ai/code/session_017GpQxEFs1Wf2CoqvVDmmfW)_


bohemian-miser:
make background selection different to icon style

bohemian-miser:
Just flagging scope before I act: this PR is narrowly the #139 mobile-title-visibility fix (header CSS in `app/lanes/page.tsx`), and "make background selection different to icon style" looks like a separate concern — the diagram's **background** selector vs the **icon style** selector (both live in `react-flow-diagram.tsx` / the toolbar via the `backgrounds` and `iconStyle` store state), not this header change. I'd rather not bolt an unrelated change onto an already-green PR.

Happy to pick it up as its own issue + branch — could you clarify what you're after so I scope it right?
- **Decouple them** (currently changing one affects the other, and you want the background choice independent of icon style)? or
- **Make the two controls visually distinct** in the toolbar (they look too similar today)?

A one-liner and I'll open a dedicated issue/PR for it.

---

PR #234, opened 2026-07-05
feat(lanes): add "Butcher's Paper" canvas theme (#111)

bohemian-miser:
Closes #111

## What & why

Issue #111 asks for a "butcher's paper" theme, wired into the existing Style toggle (the owner's triage: *"add a background/colour palette token set and wire it into the existing modern/classic mode toggle"*).

This adds **"Butcher's Paper"** as a new option in the diagram's **Style** dropdown (alongside Classic / Modern / Clean). When selected, the diagram canvas renders on a warm kraft "butcher's paper" surface instead of white:

- the diagram wrapper surface,
- the ReactFlow `<Background>` dot-pattern colour (a slightly darker grain), and
- the PNG-export background (so downloaded diagrams match the screen)

all switch together. Nodes keep the existing **classic** renderer — butcher's paper is about the *paper the recipe sits on*, so no node component is touched (dark-ink recipe on kraft paper reads cleanly).

## How it works (smallest change)

The theme decision lives in a **new pure module** `lib/recipe-lanes/canvas-theme.ts`:

- `getCanvasTheme(iconStyle)` → `{ surface, pattern, exportBackground }`. Only `'butcher'` returns the kraft palette; every other/unknown/undefined style returns the default white canvas, so **existing themes render exactly as before**.

Wiring (minimal, no store/merge logic touched):
- `react-flow-diagram.tsx` (god file — change confined to consuming the pure theme): wrapper `backgroundColor`, `<Background color>`, and both `toPng` export `backgroundColor`s now come from `getCanvasTheme(iconStyle)`.
- `app/lanes/page.tsx`: one new `<option value="butcher">` in the existing Style `<select>`.
- `types.ts`: `'butcher'` added to `IconStyleId` and the node `iconTheme` union.
- `minimal-node.tsx`: **unchanged** — `'butcher'` falls through the existing branch to `MinimalNodeClassic`, so there is zero node-rendering risk.

## How verified

- **New pure unit test** `tests/canvas-theme.test.ts` (registered in the `test:unit:pure` runner → runs in the CI **fast-checks** job): asserts `getCanvasTheme('butcher')` returns the kraft (non-white) surface; every existing style (`classic`/`modern`/`modern_clean`/`timeline-circle`) and unknown/undefined/null fall back to the default white canvas; butcher is distinct from default across all three fields; and each theme's export background matches its on-screen surface. **5/5 pass.**
[...]

PR #236, opened 2026-07-05
feat(lanes): add global "smaller leaf nodes" setting (#155)

### LINKED ISSUE #155: smaller leaf nodes as a global setting

bohemian-miser:
Closes #155

## What & why

Issue #155 asks for a **global setting to render leaf nodes smaller**. A *leaf* is a graph node with out-degree 0 — nothing consumes it (a final dish, or an ingredient that isn't used downstream). This adds an app-wide **"Small leaves"** toggle to the diagram toolbar; when on, leaf nodes render at 70% scale so the terminal nodes read as smaller/secondary.

## How it works (smallest change)

- **New pure module** `lib/recipe-lanes/leaf-nodes.ts` → `getLeafNodeIds(graph)`. Edges are implicit in this codebase (`RecipeNode.inputs` = ids that flow *into* a node), so a node is a leaf iff its id never appears in any node's `inputs`. This is the load-bearing logic and is fully unit-tested.
- **`react-flow-diagram.tsx`** (god file — change confined to the existing node-mapping loop): computes the leaf set once per layout and threads `data.isLeaf` into each node's data (applies to both `minimal` and `timeline` map paths).
- **`MinimalNode`** reads the new global `smallLeafNodes` store flag; for a leaf node it wraps its render in a uniform `scale(0.7)` transform (`transform-origin: center`). A uniform scale keeps the ReactFlow handles proportional, so edges still meet the node rather than detaching.
- **Store** (`recipe-store.ts`) gains a `smallLeafNodes: boolean` field (default `false`) + `setSmallLeafNodes` action, mirroring the existing in-memory UI prefs (`iconStyle`/`lineStyle`). It is deliberately kept **out of `graph`/`mergeSnapshot`** (per CLAUDE.md the merge path is load-bearing) because it is an app-wide UI preference, not recipe data.
- **`app/lanes/page.tsx`** (god file — one additive control): a "Small leaves" toggle button next to the existing Style dropdown.

## How verified

- **Tests added:** `tests/leaf-nodes.test.ts` (registered in the `test:unit:pure` runner) — 7 cases covering chains, multiple outputs, shared/multi-consumer nodes, isolated nodes, `undefined` inputs, empty/`null`/`undefined` graphs, and the no-edges case. All pass locally.
- Local pre-checks green: `npm run lint` (0 errors), `npm run typecheck` (clean), `npm run test:unit:pure` (full suite passes, including the new test).
- CI (`fast-checks` / `integration` / `e2e`) should exercise the same tiers — see checks on this PR.

## Scope & risks / unsure about

- **In-memory setting, not persisted.** It resets on reload, matching the existing precedent for the other UI toggles (`iconStyle`/`lineStyle` are also in-memory). If you'd prefer it to persist across reloads (localStorage), that's a small, isolated follow-up.
- **Applies to the standard "minimal" node view.** Timeline view keeps its uniform circular nodes (`data.isLeaf` is threaded there too, so wiring it up later is trivial if wanted).
- **Visual tuning:** the shrink factor is `0.7` and the scale is applied at the node component (the ReactFlow bounding box itself is unchanged, so a leaf renders smaller within its original slot rather than tightening layout spacing). I couldn't visually verify the exact look in this environment — the factor / whether to also shrink layout spacing are easy knobs to adjust in review.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FbQSfvaVteSVHE3nvhNjjF)_


bohemian-miser:
You got it around the wrong way the leaves are the ones that have no incoming edge, you changed the final node/s

also this doesn't work in timeline view

bohemian-miser:
also make it a slider

bohemian-miser:
Fixed in 5cd805b:

- **Leaf definition flipped** — a leaf is now a node with **no incoming edge** (in-degree 0, i.e. `inputs` empty — the raw ingredients), not the final nodes. `getLeafNodeIds` and its unit tests updated.
- **Now a slider** — replaced the on/off toggle with a "Leaf size" slider (40%–100%, default 100%) in the toolbar; the store holds a numeric `leafNodeScale`.
- **Timeline view** — `timeline-node` now scales leaf circles by the same factor.

---

bohemian-miser:
resolve conflicts

bohemian-miser:
Merged latest `main` (the Butcher's Paper theme, #234) and resolved the conflicts — both the new **Paper** dropdown and the **Leaf size** slider now coexist in the toolbar, and the store keeps both `canvasBackground` and `leafNodeScale`. Typecheck + unit tests pass locally; CI re-running.

---

PR #237, opened 2026-07-05
feat(lanes): add global "smaller leaf nodes" view setting

### LINKED ISSUE #155: smaller leaf nodes as a global setting

bohemian-miser:
Closes #155.

## What & why
Adds a global view setting that renders **leaf nodes smaller**. A leaf here means a node with **out-degree 0** — a terminal step that nothing else consumes (e.g. the finished dish). Edges in the graph are encoded implicitly as `inputId -> node.id` (via each node's `inputs`), so a leaf is any node whose id appears in no other node's `inputs`. This matches the interpretation in the issue's automated triage.

The setting is a toolbar toggle (**Leaves: Normal / Small**) sitting next to the existing Style / Lines / Spacing controls, mirroring the other global view settings.

## How it works (smallest-change design)
- **`lib/recipe-lanes/graph-logic.ts`** — new pure helpers `getLeafNodeIds(nodes)` and `isLeafNode(nodes, id)` (out-degree-0 detection). Pure, no React/Firestore.
- **`lib/stores/recipe-store.ts`** — new `smallerLeafNodes: boolean` state (default `false`, cleared on `reset`) plus `setSmallerLeafNodes` action, following the existing `iconStyle`/`lineStyle`/`nodeLayout` pattern.
- **`components/recipe-lanes/nodes/minimal-node.tsx`** — when the setting is on and the node is a leaf, the node is scaled uniformly (0.7) via a wrapper. This is theme-agnostic (works for both classic and modern renderers) and intentionally avoids touching the fragile layout effect in the `react-flow-diagram.tsx` god-file. Leaf-ness and the toggle are read through **boolean-returning** store selectors, so only affected nodes re-render.
- **`app/lanes/page.tsx`** — the toolbar toggle button.

The scale is a pure visual shrink: the ReactFlow layout slot keeps its size, so nodes stay put with a bit more breathing room around smaller leaves (no overlap, no re-layout needed).

## How verified
- **New pure unit tests** (`test:unit:pure`):
  - `tests/graph-logic.test.ts` — leaf detection: single terminal in a chain, multiple independent terminals, isolated node treated as a leaf, `isLeafNode` agreement, and the `undefined` guard.
  - `tests/recipe-store.test.ts` — `smallerLeafNodes` defaults to `false`, `setSmallerLeafNodes` toggles it, and `reset` restores `false`.
- Local pre-checks all green: `npm run lint` (0 errors), `npm run typecheck`, and the full `npm run test:unit:pure` (**348 pass / 0 fail**). The pre-commit hook re-ran the same and passed.

## Risks / unsure about
- **"Leaf" definition.** I used out-degree 0 (terminal outputs) per the issue's triage. If the intent was actually the raw *source* ingredients (in-degree 0), it's a one-line flip in `isLeafNode`/`getLeafNodeIds` — happy to switch.
- **Setting is not persisted** across reloads, consistent with the other in-memory view settings on this store (no `persist` middleware today).
- **Edge gap.** Because only the DOM is scaled (not the RF layout bounds), floating-edge endpoints anchor to the original slot, leaving a slightly larger gap to a shrunk leaf. Cosmetic only; the alternative (recomputing layout dimensions) would mean editing the delicate layout effect in a god-file, which I avoided deliberately.
- The scale factor (0.7) is a judgement call; easy to tune.

---
_Generated by [Claude Code](https://claude.ai/code/session_016uWwec4xoks973RXTUbWLz)_

### DIFF
--- a/recipe-lanes/app/lanes/page.tsx
+++ b/recipe-lanes/app/lanes/page.tsx
@@ -1128,7 +1128,7 @@ const handleVisualize = async () => {
                          <button
                              onClick={() => setSmallerLeafNodes(!smallerLeafNodes)}
                              className={`text-xs px-2 py-1 rounded font-medium transition-colors ${smallerLeafNodes ? 'bg-yellow-100 text-yellow-700' : 'bg-zinc-50 border border-zinc-200 text-zinc-500'}`}
-                             title="Render leaf nodes (final steps nothing else uses) smaller"
+                             title="Render leaf nodes (source ingredients with no incoming edges) smaller"
                              aria-pressed={smallerLeafNodes}
                          >
                              {smallerLeafNodes ? 'Small' : 'Normal'}
--- a/recipe-lanes/app/lanes/page.tsx
+++ b/recipe-lanes/app/lanes/page.tsx
@@ -83,7 +83,8 @@ function RecipeLanesContent() {
   const layoutMode = useRecipeStore(s => s.nodeLayout);
   const layoutModeRestoredRef = useRef(false);
   const iconTheme = useRecipeStore(s => s.iconStyle) as 'classic' | 'modern' | 'modern_clean';
-  const { setNodeLayout, setIconStyle, setLineStyle } = useRecipeStore.getState();
+  const smallerLeafNodes = useRecipeStore(s => s.smallerLeafNodes);
+  const { setNodeLayout, setIconStyle, setLineStyle, setSmallerLeafNodes } = useRecipeStore.getState();
   const [showForkPrompt, setShowForkPrompt] = useState(false);
   const [warningDismissed, setWarningDismissed] = useState(false);
   const [existingCopies, setExistingCopies] = useState<any[] | null>(null);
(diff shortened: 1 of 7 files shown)


bohemian-miser:
You got it around the wrong way the leaves are the ones that have no incoming edge, you changed the final node/s

also this doesn't work in timeline view

bohemian-miser:
Thanks — fixed both in `7c93fa1`:

1. **Leaf = in-degree 0.** `getLeafNodeIds`/`isLeafNode` now select the source nodes (no incoming edges — the raw ingredients), not the terminal nodes. Updated the pure unit tests to match.
2. **Timeline view.** The setting had no effect there because timeline uses a separate renderer (`timeline-node.tsx`); it now applies the same leaf-scale, so the toggle works in timeline mode too.

Pre-commit (lint + typecheck + pure suite, 349/349) is green; waiting on CI for the new commit.

---

bohemian-miser:
dupe of 236

PR #238, opened 2026-07-06
test(e2e): re-enable mobile device tests via a Playwright project matrix (#20)

### LINKED ISSUE #20: Re enable mobile tests
They got commented out, will probably need to fix some tests and maybe some code

bohemian-miser:
## What & why

Issue #20: the e2e mobile tests had been disabled. The Playwright config had been reduced to a single desktop Chromium project, and "mobile" coverage was faked in-spec by calling `setViewportSize()` — which only resizes the viewport and **never** exercises real device emulation (touch, `isMobile`, mobile user-agent). The owner asked for the suite to run on **both a desktop and a real mobile device via a device loop** ("I used to have a standard for loop").

This change restores a genuine mobile-device project, generated from a device list with a for-loop:

- **`playwright.config.ts`** — `projects` is now built by iterating an `E2E_DEVICES` list: a `desktop` (Desktop Chrome) project that runs the full suite, and a `mobile` project (**Pixel 5** — Chromium engine, `isMobile` + `hasTouch`) scoped via `testMatch` to the device-aware smoke suite. Adding another device is a one-line change. The mobile project stays on Chromium (not WebKit) so the existing, carefully de-flaked suite keeps its engine; scoping mobile to the smoke suite keeps the serial (`workers: 1`) CI run bounded.
- **`e2e/smoke.spec.ts`** — device fan-out is now driven by the active Playwright project instead of an in-spec viewport loop, so the mobile run happens under real device emulation. `isMobile` is derived from `test.info().project.name`. Adds a mobile-only regression guard that asserts touch is actually active (`navigator.maxTouchPoints > 0`). The two inherently desktop-only tests (banner logic, Issue-34 byline) `test.skip` on the mobile project.
- **`tests/playwright-projects.test.ts`** (new pure unit test, wired into `test:unit:pure`) — asserts the config keeps both a `desktop` project and a real-mobile project (`isMobile`, `hasTouch`, Chromium), so the mobile project can't be silently removed or degraded to a viewport-resize again — the exact regression this issue is about.

## How verified

- **New tests:**
  - `tests/playwright-projects.test.ts` (pure) — guards the device matrix; runs in the `fast-checks` job (`test:unit:pure`).
  - `e2e/smoke.spec.ts` now executes under both the `[desktop]` and `[mobile]` projects; added `mobile: real device emulation is active` asserting touch is enabled.
- **Local pre-checks (from `recipe-lanes/`):** `npm run lint` → 0 errors; `npm run typecheck` → clean; `npm run test:unit:pure` → **344/344 pass** (includes the new guard test). The pre-commit hook (lint + typecheck + unit:pure) ran and passed on commit.
- CI (`fast-checks`, `integration`, `e2e`) is expected to run these; the `e2e` job will now show smoke specs under both `[desktop]` and `[mobile]`.

## Risks / unsure about

- The smoke suite has not previously run under *real* Pixel 5 emulation (only a resized phone viewport), so the `[mobile]` project could surface a genuine mobile rendering/timing difference in CI (same ~390px width, but now with `isMobile`/touch). If a mobile assertion fails, it's a real mobile issue to fix rather than a config problem. I could not run the e2e job locally (emulators/browsers not provisioned in this environment), so CI is the first real signal.
- Scope note: mobile execution is currently scoped to the device-aware smoke suite (nav/auth/page-load), which is where the `isMobile` branching lived. Extending the interaction-heavy specs (graph drag, layout persistence, recipe lifecycle) to mobile is a follow-up that needs mobile-specific selector/responsive work verified against the live env — the config now makes adding them a one-line `testMatch` change.

Closes #20

---
_Generated by [Claude Code](https://claude.ai/code/session_015XuGQJnytDaX6Pd2JqVySg)_


bohemian-miser:
resolve conflicts and refactor so changes stop conflicting on "test:unit:pure": becasue everything hits this

bohemian-miser:
Done — both parts addressed:

**1. Conflicts resolved.** Rebased onto latest `main` (`c9a896a`); the only conflict was the `test:unit:pure` line itself.

**2. Refactored so test additions stop conflicting on `test:unit:pure`.** Pure tests are now **auto-discovered** instead of enumerated inline:
- `scripts/unit-test-manifest.mjs` — single source of truth. It lists only the emulator-dependent **integration** tests (`INTEGRATION_TESTS`); everything else under `tests/*.test.ts` is treated as a pure unit test automatically.
- `scripts/unit-tests.mjs` — `pure` runs the discovered set via node's test runner; `list-integration` feeds the emulator runner.
- `package.json` `test:unit:pure` now calls the runner (no file list), and `test-unit-integration.sh` sources its list from the same manifest.

Net effect: **adding a pure test touches no shared line**, so it can't collide there again. The change also fixed two pre-existing drift bugs the hand-maintained lists had accumulated: `autosave-jitter.test.ts` was in *neither* list (never ran — now runs, 9 tests passing), and `functions-metadata.test.ts` was double-listed in both tiers (it's emulator-dependent, so it now runs only in integration).

Local: lint 0 errors, typecheck clean, `test:unit:pure` **465/465 pass** via the new runner. Watching CI on the rebased head now.

---

bohemian-miser:
resolve conflicts

PR #243, opened 2026-07-06
fix(lanes): edge anchors follow the measured handle contract + leaf scale

bohemian-miser:
## What changed & why

Arrows have long pointed below their icons — worse for off-center ("squished") icon metadata, and worse again with the new leaf-size slider (#155 / #236). This is the systemic fix, driven by measuring the real DOM instead of trusting the code comments.

**Root cause (measured, not guessed):** ReactFlow's stylesheet (`.react-flow__handle-top { top:-4px }`) wins the cascade over the Tailwind `top-1/2` classes on the classic node's handles, so handles actually sit at the **top-center of the icon container**, not its center as `graph-utils.ts` assumed. The action path accidentally compensated (`imageY = handlePos.y` — the mystery "double compensation"), but:
- ingredient nodes were mapped over an **80px frame instead of their 56px container** → centers 12px low, metadata offsets scaled 1.43× too far (the "squished icons are worse" report),
- `getBBox` used a third, contradictory frame (handle-as-center) for ingredients,
- nodes without metadata anchored at the handle (container **top**),
- the leaf-size CSS scale was invisible to edge math entirely, and its `center center` transform origin physically dragged the icon (and every arrow ending) downward while RF's handle measurement went stale ("smaller leaves → arrows end up lower").

**Fix:** new pure module `lib/recipe-lanes/edge-anchors.ts` is the single source of truth for metadata→flow mapping, with the measured handle contract documented at the top (and a warning comment at the handle markup). `getCenter`/`getBBox`/`getRadius` all derive from one shared frame per node. `FloatingEdge` threads the leaf scale into `getEdgeParams`; `MinimalNode`'s scale origin is pinned to the handle point (per textPos), so the handle stays valid at every scale.

The metadata reference frame is the icon container — the same frame as the red-dot debug overlay previously validated by eye against painted centers.

## How it was verified

- **Browser (emulators + dev server, Playwright):** at 100% leaf size the leaf's nearest edge endpoint sits 21.5px from the icon-container center (≈ the 24px design radius); at 40% the container visually measures 22.4px and the endpoint tracks it at 8px (≈ the scaled 9.6px radius). Under the old code the anchor stayed at unscaled geometry. Screenshots taken at both scales; the 40% shot shows arrows glued to the shrunken leaf.
- **15 new pure unit tests** (`tests/edge-anchors.test.ts`, registered in the tier-0 runner) pin the frame math, the handle-point scale invariant, transform origins per textPos, and `getEdgeParams` end-to-end (bbox, scaled, and missing-metadata paths).
- `lint` (0 errors) + `typecheck` + full pure suite green via the pre-commit hook.

## Risks / unsure about

- **Deliberate behavior changes** (all more correct, but visible): ingredient anchors sit 12px higher than before; missing-metadata nodes (pending/emoji) now anchor at the icon center instead of its top edge, with radius 24 (ingredient) / 36 (action) instead of a flat 36; modern-theme *ingredient* centers now use their real 64px image frame. If any arrow looks newly off in production data, suspect these first.
- The handle contract is empirical: if anyone later "fixes" the handle CSS to actually center handles, edges will shift — the markup comment and module header both flag this, but it's a coupling worth knowing about.
- Timeline view unchanged (its edges don't use `getEdgeParams`); leaf scaling there still uses `center center` origin on a symmetric circle, which is already correct.


bohemian-miser:
Follow-up per owner review: arrows sat a constant ~6px high, most visible at small leaf sizes. Cause: RF reports a `Position.Top` handle's **top edge**, and the classic handle overhangs the container top by ~6px (DOM-measured) — frames anchored at `handlePos.y` inherited that. Fixed in 0c74080 with a documented `CLASSIC_HANDLE_TOP_OVERHANG = 6` constant; browser re-verify shows action endpoints at ~35.6px from the visual icon center (≈ the 36px design radius, previously ~29.6px = 6px high). Tests updated to pin the new expectations.

PR #244, opened 2026-07-06
feat(lanes): add Notation layout preset (spines + inline verb glyphs)

bohemian-miser:
## What & why

Adds a new **Notation** layout preset: recipes rendered as ingredient leaves
feeding thick per-vessel "spine" lines, with cooking verbs drawn as small
emoji glyphs directly in the line. Approved mockup (Panel B) is the visual
target. Purely additive — a new option in the existing Layout `<select>`
alongside Lanes/Smart/Repulsive/Timeline; no existing mode's code path is
touched.

**New files**
- `lib/recipe-lanes/verbs.ts` — pure `classifyVerb(text)` keyword classifier (chop/stir/heat/simmer/boil/drain/season/rest/fold/crush/bake/serve → emoji glyph), table-order = match priority, `null` on no match.
- `lib/recipe-lanes/layout-notation.ts` — pure `calculateNotationLayout(graph)`: one horizontal row per lane, station badge on the left, actions placed left→right by **global topological depth** (not per-lane rank — see "risks" below for why), leaves floating above the first action that consumes them, edges classified `spine` (same-lane action→action) / `drop` (leaf→action) / `cross` (cross-lane).
- `components/recipe-lanes/nodes/notation-verb-node.tsx`, `notation-station-node.tsx` — new RF node types for verb circles and station badges.
- `components/recipe-lanes/edges/notation-edge.tsx` — new RF edge type: thick round spine, thin drop line ending in a dot, rounded elbow cross edge.
- `tests/verbs.test.ts`, `tests/layout-notation.test.ts` — pure unit tests, registered in `test:unit:pure`.

**Wiring**
- `'notation'` added to `LayoutModeId` (`lib/recipe-lanes/types.ts`) and the Layout `<select>` in `app/lanes/page.tsx`.
- `react-flow-diagram.tsx`: new node/edge types registered in `INITIAL_NODE_TYPES`/`INITIAL_EDGE_TYPES`; a `mode === 'notation'` branch in `runLayout` mirrors the existing `'repulsive'` pattern (from-scratch layout, no saved-position preservation yet). Leaves/state actions reuse the existing `MinimalNode` untouched.

Leaves and existing `MinimalNode`/`calculateLayout`/`layout-force`/`edge-anchors`/`graph-utils`/store merge logic are all untouched.

## How it was verified

- `npm run lint`, `npm run typecheck`, `npm run test:unit:pure` all green (400 tests pass, pre-commit hook ran clean on every commit).
- Drove the real app in a Playwright-controlled browser (`npm run dev:emulators` from an isolated worktree, `/lanes?new=true`). Since the local `MOCK_AI` parser only ever returns a single-lane canned graph (it doesn't do real multi-lane NLP), I injected a hand-built 4-lane fixture (Pan 1 / Pan 2 / Pot / Serve) through the app's own JSON editor panel to exercise multi-lane + cross-lane rendering.
- Screenshots (kept in `~/projects/RL-notation/`, not committed to the repo):
  - `notation-before-smart.png` — same recipe in the existing "Smart" (dagre) layout, for comparison.
  - `notation-after-tall.png` — the Notation layout, full diagram: 4 station rows (PAN 1 / PAN 2 / POT / SERVE) each with a thick spine, verb glyphs sitting on the line with duration chips, ingredient leaves floating above with thin drop-lines ending in a dot, and three rounded-elbow cross-lane edges converging into the SERVE row's "Combine" step.
  - `notation-after.png` — same, standard 1440×900 viewport.
- Iterated on spacing/label-wrap twice after looking at the actual screenshots (initial pass had overlapping text and invisible cross-edges — see commit 2 for the fix and why).

## Risks / things I'm not fully sure about

- **Global-depth x positioning (commit 2):** I initially placed actions left→right by rank-within-their-own-lane (0, 1, 2…), per a literal reading of the task. That put every lane's first action — including a single-action "combine/serve" lane — at the exact same x, so cross-lane edges into it were visually indistinguishable from the column of leaf drop-lines. I switched to global topological depth for x instead, which fixed it and still satisfies "left→right in topological order," but it's a design call I made under time pressure rather than something explicitly speced — worth a second look.
- **Pre-existing timing race, not touched:** switching the Layout dropdown while a background icon-search pass is still running can silently cancel the mode-change layout's `setTimeout` (see the layout effect in `react-flow-diagram.tsx`) before it fires, because unrelated re-renders clear it first. I worked around this while testing (used the "Reset Layout" button, which calls `runLayout` synchronously) but did not fix the underlying race — it isn't specific to Notation and touching it felt out of scope for an additive PR. Flagging in case it's worth its own issue.
- **Emoji glyph rendering:** on this dev Pi's headless Chromium/Noto Color Emoji build, 🍳 (used for `cook`-type station badges) renders as a fallback glyph that looks like a magnifying glass; every other glyph in the table (🔪🌀🔥♨️🫧🫗🧂⏲️🥄🔨🍲🥘🍽️) rendered correctly in testing. This looks like a font-version gap on this specific machine, not a code bug — real browsers/OSes should render it fine — but I haven't verified on a second machine.
- **Leaf-fan crowding:** with 3+ ingredients feeding adjacent actions, sibling leaf labels can still get close to touching (tuned `LEAF_SIBLING_SPACING` to 115px against MinimalNode's ~100px label width, but didn't add collision detection across different actions' fans).
- Multi-lane testing used a hand-authored fixture rather than a real AI-parsed recipe, since the local mock parser doesn't produce multi-lane output — real Gemini output should behave the same (it's just `RecipeGraph` shape) but wasn't exercised end-to-end through the actual parser prompt.

## v2 / left for later
- Time-proportional spine segments (segment length ∝ duration).
- Station heat-color rings driven by actual `temperature` metadata rather than just lane type.
- Convergence "combine" mega-node styling distinct from a plain verb/state circle.
- Saved-position persistence for Notation mode (currently always recomputes from scratch, like Repulsive).
- The pre-existing mode-switch timing race noted above.


bohemian-miser:
**Independent pre-review completed** (4 parallel review agents over the full diff before requesting human review). Findings and dispositions:

**Fixed in `1c10695`:**
1. **Duplicate node ids for zero-input actions** (found independently by two reviewers, confirmed by reading the code): an action with no `inputs` (e.g. "Preheat the oven") is a "leaf" by `getLeafNodeIds`'s in-degree-0 definition, so the leaf-placement loop added it a second time as a floating leaf → two ReactFlow nodes sharing one id, node flickers between positions. Leaf placement now applies only to `type === 'ingredient'`; regression test added.
2. **Station badges were Backspace-deletable**: ReactFlow's default delete key removed the synthetic row anchor from canvas state (looked like data loss, set dirty flag). Now `selectable: false, deletable: false`.
3. **Save pollution**: `buildGraphForSave` wrote phantom `notation-station-<laneId>` rows into `graph.layouts.notation` on every save. Now filtered (1-line change to the shared hook; `tests/layout-saving.test.ts` still green).

**Acknowledged, deliberately not fixed in this PR** (would grow the diff; listed for visibility):
- `computeDepths` in `layout-notation.ts` duplicates the topological-ordering concern that `timeline-layout.ts`'s `topoSort` covers (different algorithm, different cycle semantics). Consolidating is a refactor across two layout engines — better as its own cleanup PR.
- `notation-verb-node.tsx` re-runs `classifyVerb` at render time rather than reading the layout-computed role from node data. They can't currently disagree (same `text` source), but it's duplicated logic; noted for a follow-up.
- The `mode === 'notation'` branches in `react-flow-diagram.tsx`'s `runLayout` follow the existing `isTimeline`/`repulsive` special-case pattern rather than introducing a layout-adapter registry. Generalizing that dispatch is the right long-term move but is exactly the kind of god-file refactor that belongs in its own PR (see `docs/architecture-review-2026-06.md`).
- New `as any` casts in `react-flow-diagram.tsx` where the three layout engines' return shapes diverge — flagged in the PR body's risk section already.
- Cycle in `inputs` silently yields depth 0 (layout misorders, no crash) — only reachable with an already-invalid graph; parser never emits cycles.

bohemian-miser:
Overseer review applied in 1e06210: the layout now guarantees every `graph.nodes` entry appears exactly once (catch-all sweep placing any node the primary passes miss — e.g. an ingredient with in-degree > 0 from an AI-adjust patch — as a 'state' node on its lane row). Without it, such a node vanished from the Notation canvas and a save would have deleted it from Firestore via buildGraphForSave's intersection. Three invariant tests added. Screenshots referenced in the body are also copied to ~/projects/RL-notation/ as stated.

bohemian-miser:
Owner-reported snap-back is fixed (2ab7134), three fixes in one commit:
1. **Snap-back root cause:** the layout effect's add/remove diff counted the synthetic station badges (RF-only nodes) as "removed from graph", so it re-ran the from-scratch notation layout on every store tick — every autosave echo or icon write snapped drags back. Stations are now excluded, and the `layouts['notation']`-arrival path (tripped by the first autosave's echo) is skipped in notation mode.
2. **Pre-existing mode-switch race fixed** (the "select says Timeline but canvas didn't change" bug): `prevMode` was committed before the 5ms layout timer, so a cancelling re-render lost the switch forever; it's now committed when the layout actually fires, so cancelled switches reschedule.
3. **Edges anchor at icon centers** for MinimalNode leaves/states (was: geometric center of icon+label, dots landed on labels).

Also merged current `main` in (picks up the #243 edge-anchor fix). Verified in-browser: mode switch applies reliably; a 128px leaf drag shows 0px drift through the autosave/echo window.

bohemian-miser:
Two more owner-reported anchor issues fixed in e19f28e:
- **Fixed left-edge offset:** the layout treated leaves/states as 56px-wide nodes, but MinimalNode renders a 100/120px wrapper with the icon centered inside — anchors landed 22px left of the icon ("about the left edge"). Layout now uses the real wrapper widths; measured drop-line start == icon center exactly (365,25).
- **Lines now track the leaf-size slider:** NotationEdge subscribes to leafNodeScale like FloatingEdge; at 40% the icon center measures (365,8.2) and the edge starts at (365,8.2).

PR #285, opened 2026-07-26
fix(timeline): guard node deletion against snapshot resurrection (#278)

### LINKED ISSUE #278: Mobile: Graph resets when deleting a node in notation view

**Likely area**:

bohemian-miser:
## What & why

Fixes #278 — "Mobile: Graph resets when deleting a node in notation view."

Deleting a node in the **Timeline (Classic)** view (`TimelineView`) removed the node with a plain `setGraph` filter and **never recorded the deletion in the store's `pendingDeletedIds` guard**. Because a node delete is not persisted to Firestore immediately, a background server write — e.g. `resolveRecipeIcons` writing an icon shortlist back — produces a snapshot that *still contains the deleted node*. `mergeSnapshot` then re-adds it and reverts unsaved state, so the graph visibly **"resets to how it was at the start."** This surfaces most readily on mobile, where `pagehide`/backgrounding timing changes when the stale snapshot is processed relative to the unsaved delete.

The `ReactFlow`/notation path (`react-flow-diagram.tsx`) already avoids this by calling `markNodeDeleted`. `TimelineView` bypassed that mechanism entirely — the asymmetry is the root cause.

### Fix (smallest change, mirrors the protected path)
`recipe-lanes/components/recipe-lanes/timeline-view.tsx`:
- `deleteNode` now routes through the store's `markNodeDeleted` (removes the node from the store graph **and** records it in `pendingDeletedIds` so `mergeSnapshot` suppresses it from incoming snapshots until the delete is persisted), plus `setDirty(true)` for parity with the ReactFlow path.
- `undo` now calls `restoreNodes` for the restored graph, clearing `pendingDeletedIds` for any node the undo brings back — otherwise an undone delete would be silently re-applied by the next snapshot.

No changes to the god files (`data-service.ts`, `app/lanes/page.tsx`, `react-flow-diagram.tsx`) or to `mergeSnapshot`'s load-bearing merge logic.

> Terminology note: the reporter said "notation". In the app, the `notation` layout renders `ReactFlowDiagram` (already guarded); the unprotected node-graph view is `timeline2` / "Timeline (Classic)" (`TimelineView`), which reproduces the exact symptom. The fix lands there.

## How verified

- **New regression test** `recipe-lanes/tests/timeline-delete.test.ts` (pure tier, auto-discovered) pins the store contract the fix relies on:
  - *BUG REPRO* — a naive `setGraph` delete is resurrected by a stale snapshot (documents the old behavior).
  - *FIX* — a `markNodeDeleted` delete survives a stale snapshot that still contains the node.
  - *FIX* — the guard clears once a snapshot confirms the delete persisted.
  - *FIX* — undo restores the node and it is **not** re-deleted by the next snapshot.
- Local pre-checks all green: `npm run lint` (0 errors), `npm run typecheck` (clean), `npm run test:unit:pure` (**480/480 pass**). The pre-commit hook (lint + typecheck + pure unit) also passed.
- CI jobs to confirm on this PR: `fast-checks`, `integration`, `e2e`.

## Risks / unsure about

- The fix achieves **parity with the notation path**: a timeline delete is now guarded against snapshot resurrection but, like the notation path, is still only persisted to Firestore on a subsequent drag/save. Persisting the delete eagerly would be a larger, separate change and is out of scope for this bug.
- The bug is platform-independent (no mobile-only code branch exists); "mobile" is where it was observed, not a distinct path. The fix corrects it on all platforms.
- The component change is exercised indirectly via the store contract (the repo's pure tier has no component-render harness — `TimelineView` pulls in `next/navigation`); the single remaining untested glue is the one-line wiring "`TimelineView` calls `markNodeDeleted`", visible in the diff.

Closes #278

---
_Generated by [Claude Code](https://claude.ai/code/session_01PUGE7u8A59ej9dyKzwsv4b)_


bohemian-miser:
dupe

PR #286, opened 2026-07-26
fix(notation): persist and protect node deletes in timeline view (#278)

### LINKED ISSUE #278: Mobile: Graph resets when deleting a node in notation view

bohemian-miser:
Closes #278

## What & why

**Reported:** "on mobile when you delete a node in notation the graph resets to how it was at the start."

The notation/timeline (`timeline2` "Classic Timeline") view's delete handler mutated **only local Zustand state**:

```ts
// timeline-view.tsx (before)
const deleteNode = useCallback((nodeId: string) => {
  pushUndo(graph);
  setGraph({ ...graph, nodes: graph.nodes.filter(n => n.id !== nodeId) });
  ...
}, [graph, setGraph, pushUndo]);
```

Two things were missing versus the ReactFlow delete path (`react-flow-diagram.tsx:275-278`), and each on its own reproduces the report:

1. **No `markNodeDeleted(id)`.** `pendingDeletedIds` is the *only* thing that stops `mergeSnapshot` from re-merging a still-present-on-server node back into the local graph. Without it, the next Firestore snapshot (an icon write-back, or the listener reconnecting) resurrected the node. Because `mergeNode` falls through to `{ ...incoming }` when a node differs structurally, that same snapshot also discarded other unsaved local state — which is why the whole graph appeared to "reset", not just the one node.
2. **No `onSave(...)`.** In this component `onSave` is the Firestore persistence path (`app/lanes/page.tsx:1276` → `saveRecipeAction`), and it was only ever called from the drag-stop handler. So the delete was never written, and a reload brought the node back.

**Why it reads as mobile-specific:** nothing in the delete path is mobile-gated. But backgrounding/locking a phone mid-edit suspends and reconnects the Firestore listener, which redelivers a fresh snapshot — the exact trigger for (1), and far more common on mobile than desktop.

### The fix

- `TimelineView.deleteNode` now mirrors the working ReactFlow path: `markNodeDeleted(id)` (suppresses resurrection until the delete is confirmed server-side) + `onSave(nextGraph)` (persists it).
- `TimelineView.undo` now calls a new store action **`unmarkNodesDeleted(ids)`** to clear the pending-delete flag for nodes it restores. Without this, undoing a delete leaves a stale flag that makes `mergeSnapshot` silently drop the restored node from every subsequent snapshot — a bug the fix would otherwise have introduced.
- Added `data-testid="node-delete"` to the timeline delete control so the e2e can select it reliably.

The store's `mergeSnapshot` / `mergeNodes` reference-preserving merge logic is untouched; `unmarkNodesDeleted` is additive and returns `{}` (stable reference) when nothing matches.

## How verified

**Added tests — both tiers.** Per `docs/WORKLOG.md` (2026-07-04/05, "Routine PR review — key learning": CI-green-but-wrong PRs that unit-tested the subsystem while user-visible behavior stayed broken), this ships a unit test for the race *and* an e2e for the real outcome.

- **`recipe-lanes/tests/notation-delete.test.ts`** (pure tier, 7 cases) — drives the store exactly as the fixed component does:
  - delete → stale snapshot still containing the node → node stays deleted (the regression);
  - the same sequence **without** `markNodeDeleted` → node is resurrected (pins the pre-fix bug so it can't silently return);
  - pending flag cleared once the server confirms the delete;
  - undo + `unmarkNodesDeleted` → restored node accepts later snapshot updates, and the negative case showing why the unmark is required;
  - `unmarkNodesDeleted` id-filtering + reference stability.
- **`recipe-lanes/e2e/notation-delete-persistence.spec.ts`** — the user-visible outcome: create a recipe, switch to `timeline2`, delete a node, wait for the auto-save POST, **hard-reload**, assert the node is still gone and the node count dropped by exactly one. Fails pre-fix (delete was never persisted).

**Local pre-checks (all green):** `npm run lint` (0 errors; my files add no new warnings), `npm run typecheck` clean, `npm run test:unit:pure` **483/483 passing**. The pre-commit hook (lint + typecheck + pure) also ran clean on the commit. Integration/e2e were left to CI per the run's convention — **see the CI status on this PR for the authoritative `fast-checks` / `integration` / `e2e` results; I have not claimed them green here.**

## Risks / unsure about

- **Delete now triggers a save.** That is the intended fix, but it makes delete a write where it previously was not, so an unauthenticated/non-owner viewer path matters. `onSave` in `page.tsx` already guards the Firestore write with `if (user && isOwner && recipeId)`, so a non-owner's delete stays local — same as before. Worth a reviewer's eye.
- **`unmarkNodesDeleted` is called with every restored node's id**, not just the ones actually deleted. That is deliberate (the timeline's undo restores a whole graph snapshot and doesn't track which node was removed) and harmless — ids that aren't pending are ignored — but it is a slightly blunter instrument than `restoreNodes`.
- **The e2e is desktop-only.** The mobile Playwright project (#238) currently matches only `smoke.spec.ts`, and this spec uses `hover()` to reveal the node controls, which has no mobile equivalent in this view. So the *persistence* half is covered on desktop and the *snapshot-resurrection* half — the actually mobile-triggered one — is covered deterministically by the unit test rather than on a real device.
- I did not touch the separate, pre-existing issue the investigation surfaced: in ReactFlow `notation` mode, verb/station nodes have no delete button, so Backspace removes them via ReactFlow's default handler without going through `handleDeleteNode` — leaving Zustand's graph out of sync. Different failure mode, out of scope here; happy to file it separately.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XjUZ7kBKpzSSujvnMRNaRT)_


bohemian-miser:
**CI status after the first run** — `fast-checks` ✅, `integration` ✅, `deploy` ✅, `e2e` ❌ (2 failures). Both test jobs executed for real (the `changes` filter resolved `code=true`; nothing skipped as docs-only).

**1. `notation-delete-persistence.spec.ts` — mine, fixed.**
The spec failed on its own precondition, not on the behavior under test:

```
expect(countBefore).toBeGreaterThan(1)
Expected: > 1   Received: 1
```

My fault: I built the recipe from a single line of free-form text, but `MockAIService` splits input by **newline** and emits one node per line — so it produced a one-node graph and the delete scenario never got to run. Switched to the `test eggs` trigger, which returns the mock's fixed three-node graph (2 Eggs / 100g Flour / Mix), and the spec now targets the terminal **Mix** node: nothing takes it as an input, so deleting it can't leave another node with a dangling `inputs` ref — keeping the assertion about *delete persistence* rather than layout recovery. Pushing that now.

**2. `graph.spec.ts:43` "Core Graph: Pan, Delete, Undo & Sidebar" — pre-existing, not from this PR.**

```
expect(locator).toContainText('2')
Expected substring: "2"   Received string: "1 Egg"
```

The identical failure — same locator, same expected/received strings — appears on **main** in runs [28932695605](https://github.com/bohemian-miser/RecipeLanes/actions/runs/28932695605) and [28926430588](https://github.com/bohemian-miser/RecipeLanes/actions/runs/28926430588) (both 2026-07-08), and the same test passed on main in runs 30100051285 and 30098375123 (2026-07-24) — i.e. it is an intermittent pre-existing failure, not a regression here. This PR does not touch the ReactFlow delete path that test exercises; the only shared code is the additive `unmarkNodesDeleted` store action, which nothing in that path calls. **I have not attempted to fix it and am not claiming it fixed** — flagging it so it isn't mistaken for fallout from this change.

The e2e log also shows the known HuggingFace `all-MiniLM-L6-v2` embedding-model download failures (`local_files_only=true` … file not found). Those are the documented non-fatal icon-search environment failures, unrelated to both of the above.

---

bohemian-miser:
**CI is green on `fc03a7c`.** All jobs executed for real — the `changes` filter resolved `code=true`, so nothing skipped as a docs-only change.

| Job | Result |
|---|---|
| `fast-checks` (lint + typecheck + `test:unit:pure`) | ✅ success |
| `integration` (emulator-backed) | ✅ success |
| `e2e` (Playwright) | ✅ success |
| `deploy` (preview) | ✅ success |

Playwright: **24 passed, 3 skipped, 0 failed**. The previous run on `e6301bb` was 22 passed / 2 failed / 3 skipped — same 27 tests, and the two failures are now both passing:

- `notation-delete-persistence.spec.ts` — my spec, fixed by switching to the mock's deterministic `test eggs` fixture. It ran and passed.
- `graph.spec.ts:43` "Core Graph: Pan, Delete, Undo & Sidebar" — passed this run without any change from me, which confirms the earlier diagnosis that it is **intermittent and pre-existing** (same failure on main in runs 28932695605 and 28926430588) rather than fallout from this PR. It may be worth its own issue; I have not touched it.

The HuggingFace `all-MiniLM-L6-v2` download errors still appear in the e2e log and remain non-fatal (mock-service fallback), as before.

Ready for review.

---

PR #288, opened 2026-07-26
fix(store): protect timeline-view node deletes from snapshot resurrection (#278)

### LINKED ISSUE #278: Mobile: Graph resets when deleting a node in notation view

bohemian-miser:
Closes #278

## What & why

> "on mobile when you delete a node in notation the graph resets to how it was at the start"

Deleting a node in the timeline/notation view reverted the whole graph to its pre-delete state.

**Root cause.** The store has an existing resurrection guard: `markNodeDeleted` records a deleted node id in `pendingDeletedIds`, and `mergeSnapshot` strips those ids out of every incoming Firestore snapshot until the server confirms the delete (`recipe-store.ts:268-282`). The ReactFlow notation path opts into this guard. **`TimelineView` did not** — its `deleteNode` (`timeline-view.tsx:332-338`) removed the node with a plain `setGraph`, so the id never entered `pendingDeletedIds`. The page's live snapshot listener (`app/lanes/page.tsx:411-442`) then re-merged the still-present node back in, and the graph appeared to "reset to how it was at the start".

**Why it reads as a mobile bug.** `TimelineView`'s only persisting path is drag-end (`timeline-view.tsx:474-483`). On desktop a user typically nudges a node after deleting, which flushes the delete to Firestore and masks the bug. On mobile a standalone tap-delete is never followed by a persisting drag, so the next server echo (e.g. an icon-forge write-back) resurrects the node reliably.

## The fix (smallest change)

- **`lib/stores/recipe-store.ts`** — new `deleteNode(nodeId)` action: removes the node, registers it in `pendingDeletedIds`, and sets `isDirty`. This mirrors what the RF notation path already gets, in one reusable place. No-ops when there is no graph.
- **`lib/stores/recipe-store.ts`** — new `restoreGraph(graph)` action: replaces the graph and **clears the pending-delete guard for every node the restored snapshot contains**. This is required by the fix, not incidental: `TimelineView`'s local undo restored via `setGraph`, so without it, undoing a delete would put the node back in the graph while it stayed in `pendingDeletedIds` — and the very next snapshot would strip it out again. Ids *not* in the restored snapshot stay pending.
- **`components/recipe-lanes/timeline-view.tsx`** — `deleteNode` routes through the store action; `undo` routes through `restoreGraph`. The now-unused `setGraph` subscription was dropped.

`mergeSnapshot` and its reference-preserving merge logic are **untouched**, as is the RF/notation delete path.

## How verified

**New pure-tier tests** — `recipe-lanes/tests/timeline-delete-resurrection.test.ts` (7 assertions across 2 suites). Pure tests are auto-discovered by `scripts/unit-test-manifest.mjs`, so this runs in the CI **fast-checks** job with no registration needed:

- **`CONTROL: a plain setGraph delete IS resurrected by the next snapshot`** — reproduces the old behavior explicitly, pinning the exact mechanism the fix addresses (this is what the timeline used to do, and the node comes back).
- `deleteNode` removes the node, registers it in `pendingDeletedIds`, and marks the graph dirty.
- a node deleted via `deleteNode` is **not** resurrected by a later snapshot that still contains it — the direct regression for #278.
- the pending id is cleared once a snapshot confirms the deletion server-side.
- `deleteNode` is a no-op with no graph.
- undo via `restoreGraph` brings the node back, clears it from the guard, and it **survives** a subsequent snapshot.
- `restoreGraph` only clears pending ids the restored snapshot actually contains.

**Local gate, green:** `npm run lint` (0 errors; the remaining warnings are pre-existing repo-wide, none in the touched files), `npm run typecheck` (clean), and the full `npm run test:unit:pure` — **483/483 pass**, run via the pre-commit hook on commit. Existing `recipe-store` (20), `timeline-layout`/`timeline-save`/`undo-complex` (34) suites all still pass.

CI: the change touches `lib/`, `components/`, and `tests/` (non-docs), so `fast-checks`, `integration`, and `e2e` should execute rather than skip as docs-only. I will confirm the run and report the outcome.

## Risks / unsure about

- **Undo semantics changed slightly:** timeline undo now clears the pending-delete guard for restored nodes. That is the intended contract (a restored node must not be re-stripped), but it means undo now writes `pendingDeletedIds` where before it only touched `graph`.
- **The delete is still not force-persisted.** This PR fixes the *resurrection* — the node no longer comes back — via the same guard the notation path relies on, and sets `isDirty` so the normal autosave path picks it up. I deliberately did not add an `onSave` call inside `deleteNode`, as that would change the timeline's persistence model beyond the reported bug.
- **No e2e/component test for the mobile tap itself.** All the decision logic lives in the store and is covered at the pure tier; the e2e tier is heavy/flaky per TESTING.md. A mobile e2e assertion would be a reasonable follow-up.
- Issue #278's title says "notation view"; the reporter's flow maps to the `TimelineView` component (the view-mode dropdown labels differ from the component names). The RF `mode="notation"` delete path was audited and is already protected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NLWD11MdVhLqiNRJz5yqbw)_


bohemian-miser:
dupe

PR #295, opened 2026-07-27
feat(lanes): tick off nodes as you cook (#281)

### LINKED ISSUE #281: Enhancement: Add ability to tick off/complete nodes
> I would like the feature to be able to tick things off as I go

URL: https://recipelanes.com/lanes?id=U7BENyEDflfW1CrSJrrZ
Reported at: 2026-07-23T23:49:17.626Z
Firestore Doc ID: feedback/ZiXwvgKWNeblmj4OWvhG

**Likely area**:
- `recipe-lanes/components/recipe-lanes/nodes/timeline-node.tsx`
- `recipe-lanes/lib/stores/` (state management for completion status)

bohemian-miser:
Closes #281

## What & why

> *"I would like the feature to be able to tick things off as I go"*

Each node now carries a `✓` toggle. Tapping it dims the node to mark that step done; tapping again undoes it.

**Design decision — completion is store state, not node state.** The tick lives in a new top-level `completedNodeIds: string[]` on the Zustand store, deliberately **not** as a field on `RecipeNode`. The save path (`data-service` → `removeUndefined(data)`) has **no field whitelist**, so any field present on a node at save time is written to Firestore verbatim. A `completed` flag on the node would therefore have:

- persisted one cook's progress and shown it to **everyone** viewing a shared recipe, and
- marked the recipe dirty on every tick, triggering autosaves.

Keeping it off the graph also means `toggleNodeCompleted` pushes no undo entry and leaves every node object reference untouched, so the per-node selectors don't re-render.

**Lifecycle.** Ticks are cleared for free by `reset()`, which `app/lanes/page.tsx` already calls in its `[recipeId]`-keyed effect (page.tsx:427) before subscribing to the new doc — so node IDs cannot bleed from one recipe into the next. An earlier draft added bespoke clearing to the store's `setRecipeId`; that action turns out to have **no app caller at all**, so the guard was dead code and was dropped.

**Mobile.** The toggle is intentionally *not* hover-gated below the `sm:` breakpoint (`opacity-100 sm:opacity-0 sm:group-hover:opacity-100`), unlike the existing reroll/forge/delete controls. Ticking off steps is a phone-in-the-kitchen action, and a `group-hover`-only control is unreachable on touch.

**Scope.** Wired into `MinimalNode` (the default node — classic + all three modern render branches) and `TimelineNode`. The issue's "likely area" suggested `timeline-node.tsx`, but that component only renders when `nodeLayout === 'timeline'`; the default layout is `dagre` → `MinimalNode`, so both are covered. Notation nodes are left alone. No god files touched (`react-flow-diagram.tsx`, `app/lanes/page.tsx`, `data-service.ts` are all untouched); the toggle is added as a sibling inside each existing relatively-positioned root, with no new wrapper `<div>`s, so ReactFlow's handle measurement and edge anchoring are unaffected.

## How verified

**Tests added** — `recipe-lanes/tests/recipe-store-completed.test.ts` (pure tier, auto-discovered by the manifest; 11 tests):

- `toggleNodeCompleted` ticks / un-ticks, tracks several nodes independently, never records a duplicate, and accepts an ID not in the graph
- **does not touch the graph or mark the recipe dirty** (asserts the `graph` reference is identical and `isDirty === false`)
- **writes no completion flag onto the node itself** — asserts the node's key set is exactly unchanged; this is the regression guard for the shared-recipe leak described above
- **does not push an undo entry**
- **ticks survive a `mergeSnapshot`** — a background write (icon forge, another device saving) must not wipe the cook's progress mid-recipe
- `reset()` clears ticks, so they cannot bleed into the next recipe

**Local pre-check:** `npm run lint` (0 errors), `npm run typecheck` (clean), and the full `npm run test:unit:pure` via the pre-commit hook — **539 tests, 539 pass, 0 fail**. No `--no-verify`.

I'll confirm the CI results (`fast-checks`, `integration`, `e2e`) once they land.

## Risks / unsure about

- **Not persisted.** Ticks are in-memory only — a page reload clears them. That felt right for a v1 (per-person, per-session state, and it keeps the change off the persistence path entirely), but if the intent is "resume mid-cook after closing the tab", that needs a follow-up using the injected-`Storage` pattern from `draft-persistence.ts`. Flagging it as a deliberate scope call, not an oversight.
- **No "clear all ticks" affordance.** Dropped it rather than ship an action with no UI caller; reload or navigating away resets progress. Easy to add if wanted.
- **Visual placement is unverified in a browser.** The toggle sits bottom-right on each node; on very small/scaled leaf nodes it may crowd the label. The dimming value (0.45) mirrors the `isDone` treatment already used in `timeline-view.tsx`. Worth an eyeball on the preview deploy — I verified structure and lint/types, not pixels.
- **Notation view** (`notation-station-node`, `notation-verb-node`) has no toggle. Deliberate scope limit; say the word if it should be there too.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cdi18b8vTgLx4AVgmZdTUw)_

### DIFF
--- a/docs/WORKLOG.md
+++ b/docs/WORKLOG.md
@@ -36,3 +36,17 @@ A shared, curated log of significant decisions, incidents, and infra/process cha
 **Testing note.** The claim logic is pure and lives in `.github/scripts/agent-claim-logic.mjs` with `node --test` coverage (including a regression case replaying the #278 collision). CI's `changes` filter deliberately excludes `.github/**` from the `code` signal, so these tests needed their own `agent-scripts` job with its own path filter — otherwise a PR touching only agent tooling would silently skip every check.
 
 **Amendment (same day).** Contract tightened after owner feedback: a worker is dispatched to **one** labeled issue, so losing the claim race means **resign** — end the run quietly as a success — not "pick a different issue". A losing run has nothing to fall back to and must not go shopping for other work to fill itself. `list`/`status` are diagnostics only (`list` now always exits 0), so exit 3 unambiguously means "resign".
+
+---
+
+## 2026-07-27 — Tick off steps while cooking (#281)
+
+**Feature.** Nodes can now be ticked off as the cook works through a recipe: a `✓` toggle on each node dims it once done. Wired into `MinimalNode` (classic + all three modern branches) and `TimelineNode`.
+
+**Design decision — completion is store state, not node state.** The tick lives in a new top-level `completedNodeIds: string[]` on the Zustand store, *not* as a flag on `RecipeNode`. This matters: the save path (`data-service.saveRecipe*` → `removeUndefined(data)`) has **no field whitelist**, so any field present on a node at save time is written to Firestore verbatim. A `completed` flag on the node would therefore have been persisted and shown one cook's progress to **everyone** viewing a shared recipe — and would have marked the recipe dirty on every tick, triggering autosaves. Keeping it off the graph also means `toggleNodeCompleted` pushes no undo entry and leaves every node object reference untouched, so the per-node selectors don't re-render.
+
+**Lifecycle.** Ticks are cleared for free by `reset()`, which `app/lanes/page.tsx` already calls in its `[recipeId]`-keyed effect before subscribing to the new doc — so node IDs cannot bleed from one recipe into the next. An earlier draft added bespoke clearing to the store's `setRecipeId`; that action turns out to have **no app caller at all** (recipe switching goes through `reset()`), so the guard was dead code and was dropped.
+
+**Mobile.** The toggle is deliberately *not* hover-gated below the `sm:` breakpoint (`opacity-100 sm:opacity-0 sm:group-hover:opacity-100`), unlike the existing reroll/forge/delete controls. Ticking off steps is a phone-in-the-kitchen action, and a `group-hover`-only control is unreachable on touch.
+
+**Tests.** `recipe-lanes/tests/recipe-store-completed.test.ts` (pure tier, auto-discovered) — toggle/untoggle, independent nodes, no-duplicate-entry, and the load-bearing guarantees: no graph mutation, `isDirty` untouched, no undo entry, **no completion field written onto the node**, ticks survive a `mergeSnapshot`, and `reset()` clears them.
--- a/recipe-lanes/components/recipe-lanes/nodes/minimal-node-classic.tsx
+++ b/recipe-lanes/components/recipe-lanes/nodes/minimal-node-classic.tsx
@@ -93,7 +93,6 @@ export const MinimalNodeClassic: React.FC<MinimalNodeViewProps> = ({
             {/* Icon Container */}
             <div
                 className={`relative ${containerSize} flex-shrink-0 flex items-center justify-center transition-all duration-200 z-10 ${selected || isPivotMode ? 'border-2 border-dashed border-blue-500 rounded-lg bg-blue-50/10' : ''} ${isPivotMode ? 'ring-2 ring-blue-400 ring-offset-2' : ''}`}
-                style={isCompleted ? { opacity: 0.45 } : undefined}
             >
                 {/* NOTE: despite the top-1/2 classes, ReactFlow's stylesheet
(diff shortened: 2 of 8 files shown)


bohemian-miser:
CI green on `595d422` — and the test jobs actually executed rather than skipping via the docs-only path filter:

| Job | Result | Duration |
|---|---|---|
| `changes` | ✅ success | 7s |
| `fast-checks` | ✅ success — Lint, Typecheck, Unit tests (pure) all ran | 28m 4s |
| `integration` | ✅ success — Build + emulator integration tests ran | 2m 38s |
| `e2e` | ✅ success — Build + Playwright ran | 4m 32s |
| `agent-scripts` | ✅ success (skipped — no `.github/scripts` changes, expected) | 3s |

No failures, nothing flaky to discount. The `fast-checks` duration is normal for this repo (the pure step runs ~27m on `main` too).

Ready for review. The open questions from the description still stand — ticks are in-memory only (a reload clears them), and the button's visual placement is worth an eyeball on the preview deploy, since I verified structure/types/tests rather than pixels.

---

bohemian-miser:
Put them in the bottom left corner of the icon instead of floating off to the bottom right

bohemian-miser:
After clicking the first tick, have a button that activates that sits with undo/redo/save etc that resets everything

bohemian-miser:
Both done in `d4d4296`.

**1. Tick moved to the icon's bottom-left corner.** You were right that it was floating — it sat at the node root's bottom-right, which in the classic/modern nodes is *below the text label*, so it drifted off the node entirely rather than sitting on the icon.

One thing worth flagging, since it changed more than a CSS class: the button was a sibling of the icon container on purpose, because CSS `opacity` on a parent can't be undone by a child — nesting it under the dimmed container would have dimmed the toggle itself, making a ticked-off step hard to un-tick. So the 0.45 dim moved off the container and onto the icon artwork: the ``, the failed-state fallback, and the emoji fallback, in each render branch. Same visual result, button stays crisp. `<Handle>` elements were not touched or reordered, so edge anchoring is unaffected.

In the timeline node the button was already a sibling of the circle body, so that one only moved left.

**2. Clear-ticks button added to the top-right toolbar**, in the same bordered group as undo/redo, next to save. It follows the undo/redo convention — always present, `disabled` until it can actually do something, i.e. until at least one node is ticked — rather than appearing and disappearing, which would shift the toolbar around mid-cook. Icon is `ListChecks`, tooltip "Clear all ticks", `data-testid="clear-completed-nodes"`.

That restores `clearCompletedNodes`, which I'd dropped from the first commit specifically because it had no UI caller. It has one now.

**Verified:** 4 new pure-tier tests for `clearCompletedNodes` (clears everything, leaves `graph` and `isDirty` alone, keeps the array reference on a no-op so subscribers don't re-render, and nodes can be re-ticked afterwards) — 15 tests in the file now. Local pre-commit ran clean: lint 0 errors, typecheck clean, full pure suite **543/543**. CI is running on `d4d4296`; I'll confirm when it lands.

Still worth your eye on the preview: I verified structure, types and tests, not pixels, so the exact corner placement at small leaf-node scales is the thing I can't check from here.

---

bohemian-miser:
resolve conflicts

bohemian-miser:
Conflicts resolved in `9d329f8` — merged `origin/main` (which had picked up #298, the `/icon_overview` pagination fix).

Only `docs/WORKLOG.md` actually conflicted: both branches appended a dated entry at the same spot. Kept both, since they document unrelated work. No code conflicts — #298 touches `shared-gallery` / `config` / `data-service`, this branch touches the node components, the recipe store and the diagram toolbar.

One nice coincidence: the #298 run hit the same `agent-claim.mjs` 401 I did and wrote it up in the WORKLOG. I appended a line noting this branch confirmed it independently, and that two runs raced #281 — the earliest claim comment won and the loser resigned, so the protocol held even with the script unusable.

Pre-commit ran clean on the merge: lint 0 errors, typecheck clean, full pure suite **555/555** (now including #298's 12 new tests). CI is running on `9d329f8`; I'll confirm when it lands.

---

PR #296, opened 2026-07-27
feat(feedback): show the feedback button on every page (#282)

### LINKED ISSUE #282: Enhancement: Show feedback button on every page
> show feedback button on every page

URL: https://recipelanes.com/lanes?id=j1htVoJzeFfAOEsWTXNK
Reported at: 2026-07-24T00:13:48.513Z
Firestore Doc ID: feedback/fNhs2HZSMkxIJvlV5qQG

**Likely area**:
- `recipe-lanes/app/layout.tsx`
- `recipe-lanes/components/feedback-modal.tsx`

bohemian-miser:
Closes #282

> **Note on history:** this PR first tried a floating "feedback pill" mounted from the root layout. The owner reported twice that it wasn't visible, and asked for the button in the top bar instead. That approach has been reverted in full and replaced with what's described below. The earlier commits remain in the history; the description below reflects the final diff.

## What & why

`FeedbackModal` is a controlled overlay with no trigger of its own, and the only thing wiring one up was `app/lanes/page.tsx` (state L81, button L962, modal L1491). Every other route — `/gallery`, `/icon_overview`, `/terms`, `/tools/umap` — had **no feedback entry point at all**.

Each page's top bar now carries a `FeedbackButton`. It sits in normal flow beside the existing nav items and reuses `navItemClass`, the same class string the Gallery / New / GitHub buttons on those pages already use, so it can't be occluded by page chrome and can't style differently from controls that are already rendering.

`FeedbackButton` owns the modal's open state, so the server-component pages (`/gallery`, `/terms`) can drop it in without client state of their own — the same pattern `LogoutButton` already uses.

**Why not a floating button:** an overlay competes with everything else pinned to the viewport. The consent banner (`fixed bottom-0 left-0 right-0 z-[90]`, full width, 121px tall on a phone) covers the entire bottom strip, so a bottom-anchored control is behind it for every user who hasn't accepted yet — i.e. every first-time visitor. A header button has no such problem.

### Two routes needed more than a top bar

Both have early returns that render **instead of** the header, so a header button alone is unreachable there:

| Location | State |
|---|---|
| `app/gallery/page.tsx:39-51` | returns a bare `<Login />` for the `mine` / `starred` / `source` filters when signed out — no header at all |
| `app/icon_overview/page.tsx:212,221` | returns a loading screen and an auth gate, both before the header |

Both gates now carry the button, so there is no page where something can go wrong with no way to report it — including trouble signing in.

`/lanes` is deliberately untouched: it already has this button, and its mobile header drops secondary items on purpose (issue #139), so I didn't re-add one there.

| File | Change |
|---|---|
| `components/feedback-button.tsx` | new — self-contained client button + modal |
| `app/gallery/page.tsx` | top bar |
| `app/icon_overview/page.tsx` | top bar + signed-out auth gate |
| `app/terms/page.tsx` | top bar (header gained a right-hand action area) |
| `app/tools/umap/page.tsx` | toolbar (light-themed, so its own colours) |
| `components/login.tsx` | button on the full-page auth gate |
| `e2e/feedback-global.spec.ts` | e2e coverage |

No god-file was touched — `app/lanes/page.tsx`, `lib/data-service.ts` and `react-flow-diagram.tsx` are untouched.

## How verified

**Tests** — `e2e/feedback-global.spec.ts`, 7 cases:

- button present and opening the modal on `/gallery`, `/gallery?filter=mine`, `/icon_overview`, `/terms` (signed out, so the latter two exercise the auth gates);
- a first-time-visitor case that clears the consent key so the banner is deliberately **up**, pinning that it no longer occludes the control;
- an authenticated case asserting the button is in the top bar of `/icon_overview` (`y < 60`), covering the header path the signed-out runs skip;
- `/lanes` still has its own header button.

Every case **clicks** the button rather than only asserting visibility. That's deliberate: the previous floating implementation passed `toBeVisible()` while being completely covered by the consent banner and unclickable. Playwright fails a click that another element intercepts, so a covered button cannot pass now.

No pure unit test: the placement logic that one covered has been deleted, and what remains is presentational.

**Manual** — Chromium against the dev server, consent pending and accepted, on `/gallery`, `/gallery?filter=mine`, `/icon_overview`, `/terms`: present and not occluded in all eight combinations.

**Local** — `npm run lint` 0 errors (258 pre-existing warnings, unchanged); `npm run typecheck` clean; pre-commit hook (lint + typecheck + `test:unit:pure`) passed on every commit, `--no-verify` never used.

## Risks / unsure about

- **`/tools/umap` is not verified.** It needs Firebase config the plain dev server lacks and throws a client-side `FirebaseError` without it, so I could not see that button render. The change there is a JSX insertion into a toolbar that does render.
- **The second invisibility report is unexplained.** The first cause (consent-banner overlap) was real and measured, but the owner still saw nothing afterwards, and I could not reach the preview URL from my environment or complete a production build locally (`prebuild` fetches a HuggingFace model; the network blocks it). There may be a remaining `next dev` vs production-build difference. The top-bar approach sidesteps it by construction — identical classes to buttons already working on the same page — but the underlying question isn't closed.
- **`components/login.tsx` is shared**, so the feedback button now appears on that gate wherever it's used, not only from `/gallery`. That seems right, but it is a wider blast radius than the page-level edits.
- **`/lanes` on mobile still has no feedback button**, unchanged from before this PR. Adding one risks re-squeezing the title that issue #139 fixed, so I left it.


bohemian-miser:
CI is green — all three required checks ran (no docs-only short-circuit; the "Skip notice" steps were themselves skipped, confirming `changes.code == 'true'`).

| Job | Result |
|---|---|
| `fast-checks` | ✅ success — lint, typecheck, `test:unit:pure` **540 tests / 540 pass / 0 fail** |
| `integration` | ✅ success |
| `e2e` | ✅ success — **27 passed, 3 skipped** |
| `agent-scripts` | ✅ success (steps skipped — no `.github/scripts/**` changes) |
| `deploy` (preview) | ✅ success |

**Confirming the new tests actually executed**, since the e2e job uses the dot reporter and prints no test titles:

- `npx playwright test --list` reports `Total: 30 tests`, matching the run's 27 + 3 exactly. The three `feedback-global.spec.ts` tests are listed under `[desktop]`.
- The 3 skips are all pre-existing and project-gated in `smoke.spec.ts`, not mine: `:59` "mobile: real device emulation is active" (`test.skip(!isMobileProject())`, skips under desktop), `:67` "Desktop: Banner Logic" and `:96` "Regression: Issue 34" (both `test.skip(isMobileProject())`, skip under mobile). That accounts for all 3, so the new specs are among the 27 passed.
- For the pure tier, the new `tests/feedback-launcher.test.ts` contributes 12 tests; the 540 total matches a local full-tier run that included it.

One note for whoever reviews: `fast-checks` took 28 minutes, essentially all of it in the `Unit tests (pure)` step (lint 11s, typecheck 7s). That is pre-existing — the new tests run in ~0.4s — but the pure tier is no longer "fast", and it also makes the pre-commit hook slow. Might be worth a separate issue.

---

bohemian-miser:
not on gallery, mine, or icon_overview what did you even add?

bohemian-miser:
You're right, and thanks for catching it — the button was there but the consent banner was sitting on top of it. Fixed in `60d4d9d`.

**What was wrong.** The launcher rested at `bottom-8` / `z-[80]`. `ConsentBanner` is `fixed bottom-0 left-0 right-0 z-[90]` and full width, so it covered the launcher completely for anyone who hadn't accepted the terms yet — which is every first-time visitor, including you on a fresh `pr-296---…` preview domain. I reproduced it against the running app: the button occupied y=653–688 while the banner occupied y=666–720, and `document.elementFromPoint` at the button's centre returned the banner, not the button. So it was both invisible and unclickable, on every page.

This was in my own "Risks" list in the PR description — I wrote that the banner covering it was "deliberate, but worth a look." That was the wrong call. A first-time visitor is exactly the person most likely to want to report something, and the banner persists until accepted.

**The fix.** The banner is 54px tall on desktop and 121px on a phone (it stacks below `sm`), so while consent is pending the launcher moves to `bottom-36` (144px) and `z-[95]`. That clears the banner outright rather than overlapping it — which also means the launcher never covers the banner's own "I agree" button, the lowest element in the stack on narrow viewports. Once consent is accepted it returns to the resting `bottom-8` / `z-[80]`.

Both components now read one shared consent store (`lib/use-consent.ts`, lifted out of `consent-banner.tsx` unchanged) rather than keeping two independent `useSyncExternalStore` copies that could disagree about whether the banner is on screen — that disagreement is the bug.

Verified in Chromium against the dev server at 1280x720 and 390x844, consent pending and accepted — uncovered in all four, and fully clear of the banner in both pending cases:

```
desktop consent=PENDING  btnBottom=576  bannerTop=666  clear=true  covered=false
phone   consent=PENDING  btnBottom=700  bannerTop=723  clear=true  covered=false
```

**Why the tests didn't catch it, which is the more useful lesson.** `e2e/utils/fixtures.ts` pre-seeds consent acceptance before every navigation, so the banner never renders in any spec — its own comment says this is precisely to stop the banner overlapping bottom-of-viewport controls. Every spec therefore ran in the one state where the bug is invisible. My e2e test also only asserted `toBeVisible()`, and a fully-covered element still passes that.

The new e2e case clears the consent key to get the real first-visit layout back, asserts the launcher sits fully *above* the banner by bounding box, and then clicks it for real — Playwright fails a click that another element intercepts, so a covered button can't pass. Plus 6 new pure cases pinning the raised offset/z-index while pending, the return to rest after, and that the `/lanes` exemption still applies while lifted.

Worth flagging: any future bottom-anchored control has this same trap waiting for it, since the fixture hides it globally.

---

bohemian-miser:
CI is green on the fix (`60d4d9d`). The preview has been redeployed, so the same URL should now show the button — including before you accept the consent banner, which was the broken case.

| Job | Result |
|---|---|
| `fast-checks` | ✅ lint, typecheck, `test:unit:pure` |
| `integration` | ✅ |
| `e2e` | ✅ **28 passed, 3 skipped** (on re-run — see below) |

`playwright test --list` reports 31 tests total, and 28 + 3 = 31, so the new first-time-visitor case did execute rather than being silently skipped. 28 passed is one more than the 27 on the previous commit, which is exactly the test I added.

**About the first e2e attempt, which I did not cause.** It failed on `graph.spec.ts:89 › Persistence Regressions: Issue 74 & 61` — `expect(locator('header h1')).toHaveText('Background Update')` timed out after 15s, with the title still reading `Mock Recipe` across 18 locator resolutions. The `onSnapshot` update never arrived at all, rather than arriving late. I re-ran the job with no code change and it passed.

I don't think this is my regression, and the reason is concrete rather than hand-waving: in the e2e environment the fixture always seeds consent, so `consentPending` is always `false`, so `feedbackLauncherClassName` emits exactly the same set of classes as before this commit (same classes, different order in the attribute — which does not affect CSS). On desktop `/lanes`, where that test runs, the launcher is `md:hidden` either way. The rendered DOM for that spec is unchanged between `6847941` (green) and `60d4d9d`, so there is no mechanism by which my diff could break a Firestore listener.

That said, I want to be straight about it rather than just labelling it "flaky": this test is **not** in the known-flakes list in `TESTING.md`, `playwright.config.ts` sets `retries: 0`, and I found no prior commits about e2e flakiness or `onSnapshot` timing. So this looks like a genuine intermittent failure in background-update propagation that hasn't been recorded before. It reproduced once in ~3 runs tonight. It may be worth its own issue — per hard rule #5 I haven't bolted an investigation onto this PR.

---

bohemian-miser:
it's still not showing on gallery and mine or icon_overview, all you should need to do is add the button to the top bar

bohemian-miser:
Done — the button is now in the top bar, and the floating pill is gone entirely (`2de5c28`).

You were right to push back on the approach. A floating overlay has to compete with everything else pinned to the viewport, and it kept losing. A header button sits in normal flow next to Gallery / New / GitHub and reuses `navItemClass` — the identical class string those already-visible buttons use — so it can't be occluded by page chrome and can't style differently from controls that are already rendering for you.

Added to the top bar of `/gallery`, `/icon_overview`, `/terms`, and the `/tools/umap` toolbar (that one is light-themed, so it gets its own colours rather than the dark nav styling). `/lanes` is untouched — it already has this button, and its mobile header deliberately drops secondary items per #139, so I left that alone rather than risk re-squeezing the title.

**Two of the pages you named needed more than a top bar**, which is worth flagging because a header button alone would not have fixed them:

- `app/gallery/page.tsx:39-51` returns a bare `<Login />` for the `mine` / `starred` / `source` filters when you're signed out — no header at all. That is the "mine" page you reported.
- `app/icon_overview/page.tsx:212,221` returns a loading screen and an auth gate, both *before* the header.

So on those, my button was in code that never rendered. Both gates now carry the button too — there shouldn't be a page left where something goes wrong and there's no way to tell me about it, including trouble signing in.

**Verified** in Chromium against the dev server, consent pending and accepted, on `/gallery`, `/gallery?filter=mine`, `/icon_overview`, `/terms` — present and unoccluded in all eight combinations.

Two things I want to be straight about rather than overclaim:

- **I could not verify `/tools/umap` locally.** It needs Firebase config the plain dev server doesn't have and throws a client-side `FirebaseError` without it. The change there is a JSX insertion into a toolbar that does render, but I haven't seen it with my own eyes.
- **I never proved why the floating pill was invisible to you the second time.** My first fix (the consent-banner overlap) was real and measured, but after it you still saw nothing, and I couldn't reach the preview URL from my environment or complete a production build locally (`prebuild` fetches a HuggingFace model and the network blocks it). So there may still be an unexplained difference between `next dev` and the deployed build. The top-bar approach sidesteps it by construction — same classes as buttons that already work on the same page — but I don't want to leave you thinking that mystery is solved.

Tests: `e2e/feedback-global.spec.ts` is now 7 cases — the button present and opening the modal on each of the four routes, a first-time-visitor case with the consent banner deliberately left up, an authenticated case pinning the button to the top bar on `/icon_overview` (the signed-out runs hit the gates instead), and `/lanes` keeping its own. Every case *clicks* the button rather than just asserting it's visible, since the last attempt passed `toBeVisible()` while being covered and unclickable. No pure unit test this time — the placement logic that one covered is deleted, and what's left is presentational.

Waiting on CI.

---

bohemian-miser:
Doesn't need the big "FEEDBACK" just the icon and mouse over

bohemian-miser:
Done — icon only now, with the name on hover (`d92bb97`). `title`/`aria-label` still carry "Feedback & Contribute", so it shows on mouse-over and is still announced to screen readers.

That also makes it match the GitHub button sitting right next to it in the same headers, which was already a bare icon with a title. The label was overpowering because the app styles all buttons with `Press Start 2P` — it renders far wider than it reads.

Applied everywhere the button appears, including the two auth-gate screens (the signed-out `/icon_overview` gate and the `<Login />` screen that `/gallery?filter=mine` falls back to), so it looks the same in all of them. Say the word if you'd rather keep a label on those two — they're standalone screens with no surrounding nav to give the icon context, so it's the one place an unlabelled icon is arguably too subtle.

Verified against the Firebase emulators before pushing: all 7 cases in `feedback-global.spec.ts` still pass. They find the button by test id and assert the modal by its heading, so dropping the label doesn't weaken them.

---

bohemian-miser:
Did you unify it so that the one on lanes page is the same <Feedback> thing as the other pages?
Make sure you've done that

bohemian-miser:
Straight answer: **no, I hadn't** — good catch. Done now in `9d7e3ae`.

`/lanes` still had its own copy: a hand-rolled `<button>`, its own `showFeedback` useState, and its own `<FeedbackModal>` render. Everything else used `<FeedbackButton />`. Two implementations of one control — which is the same drift that caused this issue in the first place, so leaving it would have been storing up the next one.

`/lanes` now uses the shared component. Its `navItemClass hidden md:flex` is passed through as `className`, so the styling and the deliberate mobile-hiding (#139) are unchanged — this swaps the implementation, not the behaviour. Net effect on `app/lanes/page.tsx` is *smaller*: the useState, the modal render and the `feedback-modal` import all go, replaced by one line. (`MessageSquare` stays imported — it's still used elsewhere in that file.)

So there's now exactly one `FeedbackButton`, used by `/lanes`, `/gallery`, `/icon_overview`, `/terms`, `/tools/umap`, and the two auth-gate screens.

**Verified** against the Firebase emulators before pushing — 8 passed, including `recipes.spec.ts` "UI Features: Feedback", the pre-existing test that clicks this exact button on `/lanes`. That was the real regression risk here, since I was replacing a control an existing test drives. `title="Feedback & Contribute"` is unchanged, so its locator still resolves.

Unrelated: the e2e run on your merge commit `4214583` went red on `graph.spec.ts` "Core Graph: Pan, Delete, Undo & Sidebar" — that's on the known-flakes list in `TESTING.md` ("screenshot can fail mid-navigation"), not something this PR touches. I didn't re-run it, because `9d7e3ae` supersedes that commit and CI is running fresh on it now.

---

PR #298, opened 2026-07-27
fix(gallery): report the real icon total so /icon_overview pagination works

### LINKED ISSUE #280: Bug: Fix pagination on icon_overview
> fix pagination on icon_overview

URL: https://recipelanes.com/lanes?id=j1htVoJzeFfAOEsWTXNK
Reported at: 2026-07-24T00:16:41.091Z
Firestore Doc ID: feedback/X5v2Eka5YI6mMsZkPzfo

**Likely area**:
- `recipe-lanes/app/icon_overview/page.tsx`
- `recipe-lanes/components/shared-gallery.tsx`

bohemian-miser:
Closes #280

## What & why

`getPagedIcons` (`recipe-lanes/lib/data-service.ts`) returned an **estimated** `total` derived from the page being requested:

```js
const totalEstimate = allIcons.length > 0 ? (page * limit) + allIcons.length : (page - 1) * limit;
```

The gallery pager turns that into a page count with `Math.ceil(total / limit)`, so the estimate always overshot by exactly one page. Concretely, with 20 icons and a page size of 20: page 1 reported `total = 40` → "Page 1 of 2" with the next arrow enabled → page 2 returned nothing, and the estimate then collapsed to `(2 - 1) * 20 = 20` → "Page 2 of 1", an empty grid, and a page number past the end of a set that now claimed to be shorter than where the user stood.

The fix returns the real size of the matching set:

- **unfiltered** — a Firestore `count()` aggregation on the *same* ordered query that pages the results, so `total` is exact and independent of `page`;
- **search** — the size of the filtered set. `icon_index` has no server-side text index, so search still scans the newest `ICON_SEARCH_SCAN_LIMIT` icons and filters in memory; only the total changes meaning.

Two smaller defects in the same function fell out of the rewrite:

- a **whitespace-only query** fell between `!query` and `query.trim()`: it took the search branch (no `offset`/`limit`) but failed the filter branch (no slice), returning up to 1000 unpaged icons. A single normalized `term` closes the gap.
- a **non-positive page** produced a negative Firestore offset. `page`/`limit` now come from a public, unauthenticated server action, so both are clamped.

The page size moves to one shared constant (`ICON_GALLERY_PAGE_SIZE` in `lib/config.ts`) used by both the client pager and the server query, so the two cannot silently drift — which is the class of bug being fixed here. `getPagedIcons` also now reads Firestore through `this._db` like the rest of `FirebaseDataService`; that is what lets the pure tier patch it.

`MemoryDataService.getPagedIcons` already returned a true total, so this brings the Firebase implementation in line with the contract the interface already implied.

## How verified

**Tests added** — both new files, covering the changed behavior directly:

- `recipe-lanes/tests/icon-gallery-pagination.test.ts` — **12 pure cases** (auto-discovered by `test:unit:pure`). Patches `service._db` with a fake chained query builder, the same technique `tests/icon-search.test.ts` already uses. Pins: total is the real collection size not a page-derived estimate; one full page does not advertise a phantom second page; a partial last page reports the full total; a page past the end still reports the real total; `page`/`limit` translate to the right `offset`/`limit`; embeddings stay stripped; whitespace-only queries are treated as no query; non-positive pages clamp to offset 0. On the search path: total counts every match, page 2 slices correctly, `count()` is *not* called (it would count the whole collection rather than the matches), matching is case-insensitive across `visualDescription` and `ingredient_name`, and a no-match search reports zero.
- `recipe-lanes/tests/icon-gallery-pagination-integration.test.ts` — **6 emulator cases**, registered in `scripts/unit-test-manifest.mjs`. This exists because a stub cannot prove the *real* query works: it seeds icons into `icon_index` under a per-run tag and asserts that `orderBy(...).count()` is supported and exact (seeding 5 moves the total by exactly 5), that the total does not vary with the page requested, that walking every page yields each document exactly once with no gaps or repeats, that an over-scrolled page is empty but still reports the real total, and that search totals count every match.

**Local pre-check** (from `recipe-lanes/`): `npm run lint` — 0 errors, no new warnings vs. `main`; `npm run typecheck` — clean; `npm run test:unit:pure` — passed via the pre-commit hook, which gates this commit.

**CI**: `fast-checks`, `integration`, and `e2e` should all execute — the diff touches `recipe-lanes/lib/**`, `recipe-lanes/components/**`, and `recipe-lanes/tests/**`, all inside the workflow's `code` path filter. I'll confirm they ran rather than skipped, and report the result on this PR.

No e2e test was added. The defect is entirely in a server function's return value, and the client's consumption of it is a single unchanged arithmetic line (`Math.ceil(total / ICON_GALLERY_PAGE_SIZE)`); the integration test exercises the real production path against the real emulator and asserts the user-visible property (page count, no phantom page, no skipped or repeated icons). Seeding 21+ icons to drive a browser pager would add meaningful e2e runtime for no additional coverage of the change. Flagging the call explicitly given the `docs/WORKLOG.md` 2026-07-04 lesson about unit-testing a subsystem while the real behavior stays broken — happy to add one if a reviewer disagrees.

## Risks / unsure about

- **`count()` support.** The unfiltered path now depends on Firestore's `count()` aggregation (firebase-admin ^13.6.0, firebase-tools ^15). The integration test is there precisely to prove this against the emulator rather than assume it. If it were unsupported, the `try/catch` would swallow it and the gallery would show an empty grid — a louder failure than today's, so worth watching the `integration` job.
- **Extra read per page load.** One count aggregation per unfiltered gallery page. Aggregations bill roughly one read per 1000 index entries, so this is negligible at current icon volumes.
- **Pre-existing, not addressed here.** `getPagedIconsAction` is public and unauthenticated, and Firestore `.offset(n)` bills `n` reads — so a caller passing a huge `page` can still run up reads. This PR clamps the low end but not the high end. It is a billing/abuse concern rather than the pagination bug, so per hard rule #5 it belongs in its own change; flagging it here so it is not lost.
- **Search totals are still bounded** by the `ICON_SEARCH_SCAN_LIMIT` (1000) scan window — unchanged behavior, but it means a search matching more than 1000 of the newest icons would still under-report. Fixing that needs a real text index, which is out of scope.

## Process note

`node .github/scripts/agent-claim.mjs claim 280` could not run in this environment: outbound HTTPS goes through an agent proxy, `GH_TOKEN`/`GITHUB_TOKEN` are proxy placeholders rather than real tokens, and direct `api.github.com` calls are refused with *"GitHub access is not enabled for this session"* — the GitHub MCP tools are the only working path. The claim protocol was honoured by hand instead: the claim comment on #280 carries the same hidden `<!-- agent-claim worker="…" -->` marker, `agent-claimed` was added, and after the settle window mine was the only live claim. `docs/WORKLOG.md` records this so the next worker does not read a 401 as "no claim needed".

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDGuwuCuUHnbG4niLPriah)_

### DIFF
--- a/docs/WORKLOG.md
+++ b/docs/WORKLOG.md
@@ -36,3 +36,13 @@ A shared, curated log of significant decisions, incidents, and infra/process cha
 **Testing note.** The claim logic is pure and lives in `.github/scripts/agent-claim-logic.mjs` with `node --test` coverage (including a regression case replaying the #278 collision). CI's `changes` filter deliberately excludes `.github/**` from the `code` signal, so these tests needed their own `agent-scripts` job with its own path filter — otherwise a PR touching only agent tooling would silently skip every check.
 
 **Amendment (same day).** Contract tightened after owner feedback: a worker is dispatched to **one** labeled issue, so losing the claim race means **resign** — end the run quietly as a success — not "pick a different issue". A losing run has nothing to fall back to and must not go shopping for other work to fill itself. `list`/`status` are diagnostics only (`list` now always exits 0), so exit 3 unambiguously means "resign".
+
+---
+
+## 2026-07-27 — `agent-claim.mjs` cannot reach the GitHub REST API from the cloud runner
+
+Working issue #280 from a Claude-Code-on-the-web session, `node .github/scripts/agent-claim.mjs claim 280` failed with **401 Bad credentials**. The cause is the sandbox, not the script: outbound HTTPS goes through an agent proxy, `GH_TOKEN`/`GITHUB_TOKEN` are `proxy-injected` placeholders rather than real tokens, and a direct call to `api.github.com` is rejected with *"GitHub access is not enabled for this session"*. In that environment the **GitHub MCP tools are the only working GitHub path** — and Node's `fetch` does not honour `HTTPS_PROXY` by default anyway, so the script bypasses the proxy even when it is configured.
+
+The claim protocol itself is fine and was honoured by hand for #280: post the claim comment with the same hidden `<!-- agent-claim worker="…" -->` marker (`add_issue_comment`), add `agent-claimed` (`issue_write`, which replaces the whole label set — resend the existing labels), wait out the settle window, then re-read the comments and keep the issue only if your claim is the earliest live one. Other workers parse the marker, not the poster, so a hand-staked claim arbitrates identically.
+
+**For future runs:** if `agent-claim.mjs` exits 1 with a 401/403, do not treat it as "no claim needed" and do not skip the claim — fall back to staking it via MCP as above. Making the script proxy-aware (an undici `ProxyAgent`, or `NODE_USE_ENV_PROXY`) would only help if the proxy allowed `api.github.com` for this session, which it does not.
--- a/recipe-lanes/components/shared-gallery.tsx
+++ b/recipe-lanes/components/shared-gallery.tsx
@@ -24,6 +24,7 @@ import { IconStats } from '@/lib/recipe-lanes/types';
 import { getIconThumbUrl } from '@/lib/recipe-lanes/model-utils';
 import { GALLERY_LABEL_TRANSFORM_CLASS } from '@/lib/recipe-lanes/gallery-label';
 import { formatIconCreatedAt } from '@/lib/recipe-lanes/icon-created-at';
+import { ICON_GALLERY_PAGE_SIZE } from '@/lib/config';
 
 interface SharedGalleryProps {
   onIconClick?: (icon: IconStats, ingredientName: string) => void;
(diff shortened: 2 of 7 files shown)


bohemian-miser:
**CI green — and the new tests provably executed, not skipped.**

| Job | Result | Duration |
|---|---|---|
| `changes` | ✅ success | — (`code` filter true, so nothing skipped as docs-only) |
| `fast-checks` | ✅ success | 27m 21s |
| `integration` | ✅ success | 2m 45s |
| `e2e` | ✅ success | 5m 02s |
| `agent-scripts` | ✅ success | 4s |
| `deploy` (preview) | ✅ success | 5m 13s |

Verified the tests ran rather than trusting a green tick:

- **pure tier** — `fast-checks` reports `# tests 540 / # pass 540 / # fail 0`. The `main` baseline (run `30266066840`) is `528`, so the delta is exactly the **+12** tests this PR adds. Individual cases confirmed in the log, e.g. `ok 1 - reports the true collection size as total, not a page-derived estimate` and `ok 6 - treats a whitespace-only query as no query` under suite `getPagedIcons — unfiltered pagination (#280)`.
- **integration tier** — all **6/6** passed under `getPagedIcons pagination (emulator)`, including `ok 1 - total reflects the real collection size via the count() aggregation`.

That last one settles the main risk flagged in the PR description: Firestore's `count()` aggregation is supported and exact against the emulator, so the unfiltered path is not silently falling into the `catch` and returning an empty gallery.

`needs-staging-lease` and `teardown` show as skipped — those are preview-lifecycle jobs and are expected to skip on a normal PR run.

---

PR #302, opened 2026-07-29
fix(undo): single store-level undo history — one Ctrl+Z, one action (#216)

### LINKED ISSUE #216: Both Ctrl+Z handlers fire on one keypress — two divergent undo systems desync graph vs positions
From the July 2026 audit (docs/architecture-audit-2026-07.md, R6 / plan item 1; PR #215).

## Problem
Two window-level Ctrl+Z listeners are registered at once:
- `app/lanes/page.tsx:98-108` → Zustand `undoStack` (graph snapshots from `setGraphWithUndo`)
- `components/recipe-lanes/react-flow-diagram.tsx:242-259` → `useHistoryManager` (ReactFlow node/edge snapshots)

When both stacks are non-empty, **one Ctrl+Z runs both undos**: the store graph pops one state while RF positions pop a different one. The two sources then disagree, and `buildGraphForSave` (which intersects them) can persist the desynced result. Redo exists only in the diagram history, so the stacks can't be re-synced by the user.

## Repro
1. Open a recipe you own on /lanes.
2. Ask the chat to adjust the recipe (populates the store undoStack via `setGraphWithUndo`).
3. Drag a node (populates the diagram history via `takeSnapshot`).
4. Press Ctrl+Z once → both the graph adjustment AND the drag revert simultaneously.

## Fix direction
Make the diagram's `useHistoryManager` the single undo owner on /lanes (remove the page-level handler and the store `undoStack` usage there), or scope one listener so only one can fire. One concern, smallest diff.

## Acceptance / proving test
e2e (Playwright, MOCK_AI): perform a chat adjustment, then a node drag, press Ctrl+Z once → exactly ONE state change reverts (the drag), and the store graph and RF node positions agree afterwards; second Ctrl+Z reverts the adjustment. Must be a browser-level test of the real outcome, not a unit test of either history in isolation.

bohemian-miser:
Closes #216.

## What changed & why
One Ctrl+Z now reverts exactly one user action. Before, **three** undo systems coexisted and two of them fired on the same keypress: the page-level Zustand `undoStack` popped a graph state (e.g. an AI adjustment) while the diagram's `useHistoryManager` popped a ReactFlow snapshot (e.g. a drag) — silently desyncing the two graph sources. Worse: once the store entry was burned, **the adjustment could never be undone from the UI**, and background icon-write snapshots resurrected any undone nodes. TimelineView held a third private stack with its own Ctrl+Z listener.

Per the view/state framework direction: **the store now owns the single history** (`undoPast`/`undoFuture` of full graph snapshots + redo); views only render and dispatch.

- **Store**: `commitNodePositions` (drag stop → one history entry), `syncNodePositions` (passive mirror of every layout pass so history always captures what's on screen), `deleteNodeWithUndo` (bridges children onto the deleted node's inputs), `pushUndoSnapshot` (layout reset / physics). Undo/redo reconcile `pendingDeletedIds` both ways so snapshots can't resurrect undone nodes, and bump `historyVersion` so views re-apply positions.
- **Diagram**: `useHistoryManager` deleted; deletes/drags go through the store; undo/redo buttons read the store; per-click history snapshots removed. `mergeNode` now treats x/y as client-owned (no identity break / no mirror clobber from background snapshots).
- **Page**: THE single window Ctrl+Z / Ctrl+Shift+Z / Ctrl+Y listener (case-insensitive key, editable-target guard so text fields keep native undo).
- **TimelineView**: private stack + key handler removed; delete/drag via store actions; `onSave` no longer rebuilds `layouts['timeline2']` from possibly-foreign node x/y.
- The #276 'hollow nodes on undo' class is now structurally impossible (history entries are graph snapshots; synthetic RF decoration never enters `graph.nodes`) — `restoreNodes` removed.

## How verified
- **Proof e2e** `e2e/undo-single-owner.spec.ts`: fails on main (after adjust+drag, one Ctrl+Z reverts both; second Ctrl+Z cannot revert the adjustment), passes here — including Ctrl+Shift+Z redo and a Firestore cross-check that no half-undone graph is persisted.
- Store history unit suite (undo/redo round-trips, pendingDeletedIds reconciliation, input bridging, sync/commit split, mergeNode identity under x/y-only diffs): 537 pure unit tests green.
- Regression e2e green locally (Pi): layout-persistence ×3 (incl. two-snapshot scenario), graph ×3, notation-delete-persistence, forge-adjust path untouched.
- An independent /code-review (multi-agent, high effort) ran before this PR; its 6 confirmed findings are fixed in the second commit.

## Risks / unsure about
- `mergeNode` no longer compares x/y (client-owned). Mid-session position changes from another device won't move nodes locally — but they already didn't (the diagram deliberately ignores position-only snapshot changes); positions still load from `graph.layouts` on open.
- Undo in notation mode recomputes the spine layout (saved-position restore for notation lands in the next PR of this series, #268).
- TimelineView undo now persists the restored state via its existing `onSave` path — same immediacy as its delete persistence (#286), new for undo.

https://claude.ai/code/session_01M9cMcYAKjNSMjRHtYSM5av


bohemian-miser:
**Update — notation undo/redo snap-back fixed** (owner repro: "undo 3 things and redo resets all 3, always back to original").

Root cause: notation mode never applied saved positions on restore (`runLayout`'s `isNotation` branch always recomputed the spine layout from scratch — a pre-existing 'v2 concern'). This branch made it visible on every undo/redo, because restoring a graph now re-runs layout instead of restoring raw ReactFlow snapshots. The new commit overlays `graph.layouts['notation']` onto the computed layout, which fixes undo/redo positions in notation AND the long-standing #268 snap-back on reload/mode-switch.

Proof: `e2e/notation-layout-persistence.spec.ts` — 272px snap-back on main, <10px restore here; full regression sweep green locally (notation×2, undo, layout-persistence×3, 564 unit tests).

PR #309, opened 2026-08-03
fix(save): buildGraphForSave unions store nodes with rendered nodes (#222)

### LINKED ISSUE #222: buildGraphForSave drops store nodes not yet rendered in ReactFlow — autosave can silently delete nodes
From the July 2026 audit (docs/architecture-audit-2026-07.md, R4 / plan item 8; PR #215).

## Problem
`components/recipe-lanes/hooks/useSaveAndFork.ts:39-41`: the saved node list is `graph.nodes.filter(n => currentNodes.some(rn => rn.id === n.id))` — the **intersection** of store nodes and currently-rendered RF nodes, with `inputs` rebuilt from RF edges only. A store node that hasn't reached RF yet (snapshot-merged node from `addNodeToRecipe`/another tab, or an AI-adjust node during the 5ms layout `setTimeout` window in `react-flow-diagram.tsx:440-444`) is dropped from the save → **deleted in Firestore** by the next debounced autosave. Corruption is silent and permanent.

## Fix direction
Save the union: nodes present in the store but absent from RF keep their store `x/y/inputs` verbatim instead of being filtered out. Note the header comment in `buildGraphForSave` — the `graph.layouts` mutation is intentionally pinned by `tests/layout-saving.test.ts`; don't "fix" that as a drive-by.

## Acceptance / proving test
Unit (the helper is pure, extracted for exactly this): store graph has node X absent from rfNodes → returned graph still contains X with its store position and inputs; nodes present in both still take RF positions. Plus an emulator/e2e check: while a recipe is open, add a node via a second write path (e.g. `addNodeToRecipe`), immediately drag an existing node → after autosave, Firestore still contains the added node.

bohemian-miser:
Closes #222. Fourth PR in the framework-first remediation (after #302 undo, #304 field-ownership map, #305 save reconciliation).

## The bug
`buildGraphForSave` decided which nodes to save by **intersecting** the store's graph with the nodes ReactFlow currently has rendered:

```js
graph.nodes.filter(n => currentNodes.some(rn => rn.id === n.id))
```

Any node in the store but not currently rendered — e.g. one just added by an in-flight adjust, or not yet mounted — was silently dropped from the saved graph. Autosave then deleted it from Firestore. Losing user data to a rendering race.

## The fix
Make it a **union**: map over `graph.nodes` instead of filtering it.

- Rendered node → unchanged behaviour: takes its ReactFlow position and `rfEdges`-derived `inputs`.
- Unrendered node → passes through untouched, keeping its store `x`/`y`/`inputs`. (Deliberately *not* re-deriving `inputs` from `rfEdges` — there are no edges for an unrendered node, so that would wrongly blank them.)

Untouched on purpose: `layouts[mode]` still contains only rendered nodes (an unrendered node has no rendered position to record), the synthetic `lane`/`notation-station` exclusion, and the pinned `graph.layouts` handling the function's NOTE warns against "fixing".

## Verification
New unit tests in `tests/layout-saving.test.ts` (which already covers this helper):
- An unrendered store node survives with its original `x`, `y`, and `inputs`.
- A rendered node still gets its RF position and `rfEdges`-derived inputs.
- `layouts[mode]` contains only the rendered nodes — the unrendered one must not appear.
- Node ordering matches `graph.nodes`.

Full pre-commit gate green, no `--no-verify`: lint + typecheck clean, **`test:unit:pure` 585/585**.

## Risks
Low. The change strictly *adds* nodes to the saved graph that were previously dropped; no node that was saved before is saved differently. The one judgement call is keeping an unrendered node's stored `inputs` rather than recomputing them — correct, since `rfEdges` has no information about a node that isn't on the canvas.

### DIFF
--- a/recipe-lanes/components/recipe-lanes/hooks/useSaveAndFork.ts
+++ b/recipe-lanes/components/recipe-lanes/hooks/useSaveAndFork.ts
@@ -41,15 +41,16 @@ export function buildGraphForSave(
     const layouts = { ...(graph.layouts || {}) };
     layouts[mode] = currentNodes.map((n: any) => ({ id: n.id, x: n.position.x, y: n.position.y }));
 
-    const nodesWithPos = graph.nodes
-        .filter(n => currentNodes.some((rn: any) => rn.id === n.id))
-        .map(n => {
-            const rfn = currentNodes.find((rn: any) => rn.id === n.id)!;
-            const inputs = rfEdges
-                .filter((e: any) => e.target === n.id)
-                .map((e: any) => e.source);
-            return { ...n, x: rfn.position.x, y: rfn.position.y, inputs };
-        });
+    // Union, not intersect: a store node not currently rendered (in-flight
+    // adjust, not yet mounted) must survive the save unchanged, not be dropped.
+    const nodesWithPos = graph.nodes.map(n => {
+        const rfn = currentNodes.find((rn: any) => rn.id === n.id);
+        if (!rfn) return n;
+        const inputs = rfEdges
+            .filter((e: any) => e.target === n.id)
+            .map((e: any) => e.source);
+        return { ...n, x: rfn.position.x, y: rfn.position.y, inputs };
+    });
 
     return { ...graph, nodes: nodesWithPos, layouts, layoutMode: mode };
 }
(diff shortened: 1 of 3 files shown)


bohemian-miser:
I made an adjustment and it shows the adjustmented recipe and then overrides it immediately. Pls fix

PR #321, opened 2026-09-13
Gallery: multi-recipe ingredient comparison table (signed-in only)

bohemian-miser:
## Summary

Signed-in users can tick several recipe cards in the gallery (Gallery / Mine / Starred tabs) and get a comparison table at the top of the page:

- **Columns** are the selected recipes (preview icon + title, linking to the editor), with a remove button per column.
- **Rows** are ingredients (icon, label, unit on the left), one row per ingredient + unit pair.
- **Cells** hold each recipe's quantity; a **Total** column on the right sums matching units across all selected recipes.
- **Rows and columns are drag-and-drop reorderable** (native HTML5 DnD, no new dependency), plus arrow keys on the grip handles as a keyboard fallback.
- Guests see neither the tick boxes nor the table.

## How it works

- `lib/recipe-lanes/comparison-table.ts` (pure): rows are keyed by the app's standardized ingredient name **plus unit** (summing "200 g flour" with "1 cup flour" would be meaningless, so units split rows); repeated lines inside one recipe are summed; quantities are scaled to the recipe's current `serves` exactly like `handleUpdateServes` in the editor; a line without a number shows `✓` and marks the row total as partial (`5+`) instead of silently reading as zero. `reconcileOrder` keeps a user-arranged order stable as recipes are added/removed.
- `getComparisonRecipesAction(ids)`: login required, capped at 12 recipes, returns the slim `ComparisonRecipe` shape (icon shortlists never cross the wire). It runs on the Admin SDK, which bypasses rules, so `canViewRecipeForComparison` re-applies the Firestore `recipes` read rule (public/unlisted, or owner) per id. Unreadable ids are dropped and the client unticks them with a notice.
- Selection state (`components/recipe-comparison/comparison-context.tsx`) is mirrored to `sessionStorage` per user through `useSyncExternalStore`, so it survives switching gallery tabs (the page is server-rendered and remounts) without a hydration mismatch. The provider is mounted by `app/gallery/page.tsx` only when a session exists.

## Tests

- `tests/comparison-table.test.ts` (pure tier, auto-discovered, 26 cases): extraction, serves scaling, unit-split rows, within-recipe summing, partial totals, float noise, order reconciliation, `moveItem`, access gate.
- `e2e/recipe-comparison.spec.ts`: creates two mock recipes, ticks both, asserts columns/rows/cells/totals, reorders rows and columns by both drag and keyboard, selection survives a tab switch, remove/clear, and a guest context sees no toggles on a published card (new admin helper `publishRecipe`).
- Verified locally: lint (0 errors), typecheck, full pure tier (626/626), and the new e2e spec against fresh emulators.

## Screenshot

Two mock recipes compared on the Mine tab (Eggs 2 + 2 = 4, Flour 100 g + 100 g = 200 g, Butter present in one recipe without a quantity):

Columns: recipe icon + title with grip and remove controls · Rows: ingredient icon, label, unit · Right: yellow Total column.

## Notes

- Also appends a WORKLOG entry documenting the design decisions.

https://claude.ai/code/session_01GohM5ykYA2KKFSAo1XJY43

---
_Generated by [Claude Code](https://claude.ai/code/session_01GohM5ykYA2KKFSAo1XJY43)_


bohemian-miser:
**CI note: `e2e` failed on `fe40ea5` in a test this PR does not touch.**

- Failing test: `e2e/graph.spec.ts` › *Core Graph: Pan, Delete, Undo & Sidebar* — after clicking the sidebar `+` (serves stepper) the Egg node still read `1 Egg` instead of `2 Egg` for 10 s. 34 other tests passed, including `recipe-comparison.spec.ts`.
- Why it isn't this PR's: the commit changed only the gallery comparison files (`components/recipe-comparison/*`, `lib/recipe-lanes/comparison-table.ts`, its unit test and e2e spec, WORKLOG). Nothing in the editor, sidebar, or serves scaling. The same `e2e` job was green on the previous head `3997376`, and `TESTING.md` §3 lists exactly this test under *Known flakes*.
- No fix PR exists for the flake, so nothing to port. I re-ran the failed job once (`rerun_failed_jobs`).

Proposed hardening for a separate PR: in `graph.spec.ts` wait for the serves stepper to render its current value before clicking `+` (e.g. `await expect(sidebar.getByText('1', { exact: true })).toBeVisible()`), and assert the node text with a `toHaveText(/2 Egg/)` after the sidebar value itself reads `2`, so the click can't land before the sidebar's handler is wired.

---

